Issue #146, opened 2025-05-09
🐞 Bug: .env file not recognized


donslice:
### Bug Description

Created a stack using the `.env` feature. Starting stack results in 500 Error.

### Steps To Reproduce

1. Create Stack with compose.yaml
2. Ensure `.env_file` is set.
3. Create .env configuration
4. Save
5. Start

### Expected Behavior

Stack starts with no errors

### Actual Behavior

Instead, I get a 500 Error.

Browser Dev console shows: `{ success: false, error: "Failed to start stack: ENOENT: no such file or directory, open '.env'" }`

### Screenshots

![Image](https://github.com/user-attachments/assets/cde0c17e-4783-4ebe-8a36-d6b07a7d323c)

![Image](https://github.com/user-attachments/assets/9a60597a-4b81-48f3-af71-dca601257750)

### Arcane Version

v.0.7.0

### Operating System

Linux

### Browser

Firefox 138

### Docker Version

26.1.4

### Additional Context

Arcane Compose

```
services:
  arcane:
    image: ghcr.io/ofkm/arcane:latest
    container_name: arcane
    hostname: arcane
    user: root
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /opt/arcane:/app/data
    networks:
      - proxynet
    environment:
      - PUID=1000
      - PGID=1000
      - APP_ENV=production
      - DOCKER_GID=998
      - PUBLIC_SESSION_SECRET=
      - PUBLIC_ALLOW_INSECURE_COOKIES=false
    restart: unless-stopped

networks:
  proxynet:
    name: proxynet
    external: true

```

donslice:
Using the full .env path of `/app/data/stacks/beszel/.env` works, but is not convenient.

kmendell:
Huh interesting, Thanks for reporting, Ill take a look at this first thing tomorrow

kmendell:
@donslice Can you try this image? `ghcr.io/ofkm/arcane:env-fix` to see if this is fixed?

donslice:
@kmendell Changed `env_file` back to `.env` and started -- Did not work.

```
arcane  | Entrypoint: Group with GID 1000 already exists: arcane
arcane  | Entrypoint: User with UID 1000 already exists: arcane
arcane  | Docker socket accessible at /var/run/docker.sock (GID: 998)
arcane  | Entrypoint: Setting permissions on critical directories...
arcane  | Starting Arcane as user arcane (1000:1000)...
[...]

PR #153, opened 2025-05-09
fix: env not being loaded for stacks

kmendell:
Fixes: https://github.com/ofkm/arcane/issues/146

## Summary by CodeRabbit

- **New Features**
  - Environment variables from `.env` files are now automatically injected into Docker Compose services when starting, restarting, or redeploying stacks.

- **Chores**
  - Added a new dependency to improve environment variable handling.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "codemirror": "^6.0.1",
         "dockerode": "^4.0.6",
         "dockerode-compose": "^1.4.0",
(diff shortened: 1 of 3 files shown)


kmendell:
This shouldnt be needed, i found a simplier solution in: https://github.com/ofkm/arcane/commit/7090c4e0950274e0334bc229e5c3b1435ee3e22d

PR #401, opened 2025-09-08
refactor: remote environments with access tokens

kmendell:
This should not be tested in the current state in a Prod environment, this still is a WIP


kmendell:
@coderabbitai review

Issue #448, opened 2025-09-12
⚡️ Feature: Improve UX


pedrofmp:
### Feature Description

Retain the view setting once i change it. On the container view, once i remove the column image, it would be nice to  keep this setting through sessions.
As for the Environments, it would be nice to refresh the information once i select a new environment.

### Problem It Solves

Better UX usability 

### Proposed Solution

Save the information on localstorage ( cookies)

### Would you be willing to work on this feature?

- [ ] Yes, I'd like to implement this feature
- [ ] I could help with parts of this feature
- [ ] No, I'm just suggesting the feature

kmendell:
Yes this is something i am pondering but due to the way i redid the tables i just didnt have a great idea on how to do it. -- as far as the environments it should refresh once you select one, (minus the dashboard as there is still a bug where it doesnt show the correct information.) Are you not seeing this?

pedrofmp:
For the environment, it refreshes if i do manually, i was just wondering if this is something that could be done automatically no biggie though. TBH, i think the view issue is more important because some of the images have a long name and it hides the 3 action dots at the end.
This is one example of such names: 
ghcr.io/immich-app/postgres:14-vectorchord0.4.3-pgvectors0.2.0@sha256:32324a2f41df5de9efe1af166b7008c3f55646f8d0e00d9550c16c9822366b4a

kmendell:
Ahh yeah skipped my mind if a pined sha image is used... I will see what i can do. :)

pedrofmp:
i appreciate it. portainer is already out of the picture. and this will only make arcane better !

PR #449, opened 2025-09-12
feat: persistent table filters, page size, and column headers

kmendell:
Fixes: https://github.com/ofkm/arcane/issues/448


user00265 (approved):
Tested on my install and nothing looks horribly out of place. Good to go!

Issue #440, opened 2025-09-11
🐞 Bug: Unable to use local Templates


pseudomix:
### Bug Description

Hello,
I'm trying to use te local templates placing my files in /app/data/templates/compose folder but they never appear in the GUI.
How can i check what I'm doing wrong ?

### Steps To Reproduce

- Place files in /app/data/templates/compose
- Not displaying in the GUI

### Expected Behavior

Expect to see the templates int he GUI when i place for exemple:
- base.yml
- base.env

in the /app/data/templates/compose

### Actual Behavior

I checked the access rights (arcane:arcane 600 for files)
I try to store files in this two differents ways:  
```
  volumes:
      - ./data:/app/data
      or
      - arcane-data:/app/data
```
Is the /app/data/templates/compose is the good path ?

I check the logs but no error.

Have a nice day

### Arcane Version

v1.0.1

### Operating System

Linux

### Browser

Brave

### Docker Version

25.0.2

jux-stef:
Same issue here ✋

kmendell:
Thanks for reporting i will look into this.

kmendell:
Okay this will work in the next release however, the structure has changed a bit to make sure it aligns with the registry templates. 

data/templates is the directory. 

The structure looks like:

data/templates
---pocketid/compose.yaml
---pocketid/.env.example 

This would then get picked up by the template service and imported as a local template.

pseudomix:
Great! Thanks kmendell for the correction and this great project !

PR #462, opened 2025-09-13
fix: allow use of the local templates directory

kmendell:
Fixes: https://github.com/ofkm/arcane/issues/440


kmendell:
@coderabbitai review

PR #511, opened 2025-09-17
chore: Auto close VSCode terminals when tasks are finished

cabaucom376:
I accidentally disabled the auto-closing feature of the terminals after completing their specific tasks. (I mistakenly believed it prevented error states from displaying, but it wasn’t.) This can clutter up your integrated terminal if you execute multiple tasks, and this addresses that issue.


kmendell:
Gotcha, Thanks!

PR #529, opened 2025-09-20
feat: support PWA icons

cabaucom376:
Part of improving mobile support.


kmendell:
We should use the app.webmanifest for PWA support, as this seems to be the standard going forward https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest

kmendell:
I havent looked at the rest yet but that should be a simple change. Ill try to look at the rest as soon as i can.

cabaucom376:
I’m not sure if it’s just some weird caching in my iOS simulator or what, but with the current "next" build the status bar safe area is the proper background color but on my dev build I am just getting plain black behind the status bar.

kmendell:
@cabaucom376 From the chnages in this PR or seomthing different? I will look at this tomorrow, got back alot later than i expected.

kmendell (approved):

PR #537, opened 2025-09-21
WIP feat: mobile navigation

cabaucom376:


cabaucom376:
I’m not sure how to achieve this, but I’d like to consolidate the logic between the floating and docked components into a single file. This way, I can edit the logic and it will be applied to both. Also the physics behind dragging the sheet to close needs more refinement. And the last feature I can think of is some sort of first use hint/guide for how to open/trigger the nav sheet.

kmendell:
I can try to take a look later tonight at consolidating it. just because the amount of chnages ill have coderabbit review it as a "prelim" code review.

kmendell:
@coderabbitai review

frontend/src/lib/utils/persisted-state.ts
kmendell, line 1: I think an easier solution vs having the frontend handle all of this would be to save it to the user in the database, and then it all gets populated with the user store automatcially once the type is updated in the frontend.

frontend/src/lib/stores/config-store.ts
```diff
@@ -13,8 +13,14 @@ const set = (settings: Settings) => {
 	settingsStore.set(settings);
 };
 
+
+const update = (updater: (settings: Settings) => Settings) => {
```
kmendell: Why the need for this update function? you can just use set() and then reload() you just have to spread the exissting settings as well ie: `...settings`

backend/resources/migrations/sqlite/016_add_navigation_settings.up.sql
kmendell, line 1: The migrations shouldnt be needed, you can just use the settings_service.go to define those default values

kmendell:
@cabaucom376 let me know when you think this one is fully ready, ill do the proper review on it.

kmendell:
@cabaucom376 The PR i just merged was kinda a beast so it may have caused a decent amount of issues.... sorry about that. I can look at the issues once i get home.

cabaucom376:
I see that now lmao

frontend/static/app.webmanifest
```diff
@@ -2,8 +2,8 @@
   "name": "Arcane",
   "short_name": "Arcane",
   "description": "Modern Docker container management platform",
-  "theme_color": "#000000",
-  "background_color": "#ffffff",
+  "theme_color": "oklch(0.141 0.005 285.823)",
+  "background_color": "oklch(0.141 0.005 285.823)",
```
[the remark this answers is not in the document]
kmendell: According to the spec its can parse them fine https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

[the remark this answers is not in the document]
kmendell: according to the spec its fine https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

cabaucom376:
Way cleaner and all amazing improvements! Yes I dealt with that issue before and ended up just adding a small time delay that would allow you to transition the cursor over the pop-up which would then also have hover detection. Feels slightly hacky but it worked. (edit: nevermind I see what bug youre talking about)

cabaucom376:
@kmendell just submitted a fix for that specific issue.

PR #563, opened 2025-09-24
fix: correct display of user label in event details

neilsb:
When viewing details of an event in the event log, the **user** label was displaying the function code rather the localised text due to the missing paratheses on `m.common_user()`


kmendell (approved):

kmendell:
Thank you!

kmendell:
Ill fix the playwright issue in a bit.

PR #573, opened 2025-09-25
fix: refresh container list when environment changes

neilsb:
Addresses a couple of issues on the Container List page.    The same issues exist on the projects, images, networks and volumes screens, and I am happy to make similar changes on them - but thought it best to check this approach is acceptable first.

### 1. Refresh Container List data as soon as new environment is select
Before, if you were on the container list screen and selected a new environment from the dropdown, the container list did not automatically update until the refresh button was pressed or the page was reloaded.

Now when the environment in the environment store is changed, `RefreshContainers()` is called.

### 2. Container Counts updated
Before, the container counts were not updated when the refresh button was pressed, meaning if containers were stopped/started the refresh button wouldn't show these changes.  Additionally, if the environment was changed and refresh was pressed it would show containers from the new environment, but leave the container counts from the original environment showing.

Now `getContainerStatusCounts`() is also called as part of the refresh.


kmendell:
Yeah this approach is accectable to me, Feel free to make thise changes on the other pages as well. Id like to make use of that handleApiReult with callbacks function if possible but it only accepts one promise at a time so the code can get a little "bulky" for lack of a better term.

neilsb:
No problem, I'll switch to using 2 separate calls to that then.   Do you want the changes to all pages in a single PR, or would you prefer one for each?

kmendell:
You can make them all in this one that works for me :)

neilsb:
Updates made to all mentioned pages now.  Not totally keen on the changes required to track the refresh status when doing 2 call 2 calls to `handleApiResultWithCallbacks()` but couldn't think of a neater way offhand.

Also replaced few hard coded strings that were already defined in the localization files while in those screens.

kmendell:
@neilsb The reason id rather keep that is that way once i get around to it, that can extract the actual error messages (like its supposed to), I do agree its not the best... if you have ideas on how to make that helper better im all for it as well. 

Otherwise this looks good to me.

Issue #310, opened 2025-06-20
⚡️ Feature: Global Environment Variables


txwgnd:
### Feature Description

Hello All,
would be great to see a feature that implements global environment variables. 
These variables are getting used in conjunction with the per stack variables found in .env.

### Problem It Solves

From my own homelab environment I know that there a a lot of values that are used in more than one stack.
It would be very convenient to save them as environtment variables to use them in all stacks and be able to change them in one place rather than many. Good examples for this are domains, paths and ports.

### Proposed Solution

Add a .global.env file in the root Stacks directory and invoke it during startup of stacks.

.env			- Found in every stack folder with variables for that stack
.global.env	- Found in the Stacks directory with variables that are commonly needed for a lot of stacks e.g. domain

### Would you be willing to work on this feature?

- [ ] Yes, I'd like to implement this feature
- [ ] I could help with parts of this feature
- [x] No, I'm just suggesting the feature

kmendell:
Ill try to add this in 1.0, but first and foremost i want to make sure 1.0 is stable in the basic features first.

txwgnd:
Legends! @kmendell & @cabaucom376 
Cannot thank you enough for this. ♥️

PR #620, opened 2025-10-03
feat: global environment variables

cabaucom376:
Global environment variables, local variables have priority and override global ones. I believe this is what #310 was actually referring to.


kmendell:
The docker compose api can pass the system environment varibles into the container, i may need to enable it though. I think thats a easier solution vs making a complete seperate logic for passing env in dont you think?

cabaucom376:
That’s fine, but the way I use Docker, I prefer being able to copy a single folder and have a fully reproducible setup. That’s why I stick to local bind mounts instead of volumes. In the same vein, having a global env file makes sense when parameters are shared across multiple compose files. This would still keep everything contained within the project and not rely on the proper system variables being in place in case you have to migrate systems.

Also, this approach has the downside of not being able to down a project if you remove one of the global variables while it’s running. Due to the compose needing to be loaded for it to bring it down. I had a workaround implemented in which, as a fallback, the project was brought down by name, and it solved said issue, but I figured you might think of something I didn’t.

PR #638, opened 2025-10-05
feat: proper docker error messages in toasts

kmendell:


kmendell:
i messed up somewhere so going to start over.

PR #703, opened 2025-10-13
chore: simplify i18n

cabaucom376:
I am not sure if this is a change that is even desired, but I just noticed that there was a lot of i18n overhead when it comes to lots of things that could be consolidated into more generic patterns for easier maintenance and less resource-specific keys.


kmendell:
Oh this is 100% something i wanted to do , i just could never find the time too. Let me know if you would like me to help with this.

cabaucom376:
@kmendell, this one is going to be a pain with merge conflicts, so it’s probably best to sort out and merge after the templates consolidation is complete. I am also not really sure how Crowdin works, so I am only touching the en.json file, but I have categorized all the keys to help find them easier. I am not really sure how that will be handled with the other languages.

kmendell:
@cabaucom376 Can we merge this into main instead of the templates branch? Id assume we can just git cherry-pick the rnage of commits here and then make a new branch for it with out those changes. That way we can avoid an annoying Merge commit and keep the audit log for what changed

cabaucom376:
@kmendell feel free to do that, I know I added some i18n strings in the templates branch which is the only reason why I build off of that so that I had the latest ones. Alternatively we could just merge the templates branch and swap to base branch to main.

PR #705, opened 2025-10-13
feat: update scheduling

cabaucom376:
**Disclaimer** Greptiles Reviews uses AI, make sure to check over its work

<h2>Greptile Overview</h2>

Updated On: 2025-10-28 20:27:16 UTC

<details><summary><h3>Greptile Summary</h3></summary>

Implements flexible update scheduling system with per-window timezone support, replacing the simple autoUpdateInterval with configurable maintenance windows. Projects can now override global settings to define custom update schedules.

**Key Changes:**
- Migrates from `autoUpdateInterval` to `updateScheduleEnabled` + `updateScheduleWindows` configuration
- Each schedule window has its own timezone, allowing multi-region scheduling
- Projects can override global update settings with per-project schedules
- Migration converts existing `auto_update = false` to NULL for settings inheritance
- Adds UI components for configuring schedules and managing project overrides

The implementation addresses previous critical issues: `$state` declarations fixed in onboarding, JSON struct tags added to UpdateScheduleWindow, and form reactivity corrected in ProjectSettings.svelte.

</details>
<details><summary><h3>Confidence Score: 5/5</h3></summary>

- This PR is safe to merge with all critical issues from previous reviews resolved
- All previously identified critical issues have been addressed: UpdateScheduleWindow now has proper JSON struct tags, ProjectSettings.svelte correctly uses $derived with createForm (not recreating form on each update), settings_dto.go correctly implements the per-window timezone approach (no global UpdateScheduleTimezone field needed), and image_polling_job.go uses global settings which is a valid implementation choice. The migration properly handles the false-to-NULL conversion for inheritance semantics.
- No files require special attention - all previously flagged issues have been resolved
</details>

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| backend/internal/dto/settings_dto.go | 5/5 | Adds UpdateScheduleEnabled and UpdateScheduleWindows fields to UpdateSettingsDto, correctly matching the migration schema |
| frontend/src/routes/projects/components/ProjectSettings.svelte | 5/5 | Project settings override UI with form handling - form is correctly created using $derived with createForm, globalSettingsDescription properly wrapped as function |
| backend/internal/services/settings_service.go | 5/5 | Adds ParseUpdateScheduleWindows and ValidateUpdateScheduleWindows methods with proper validation for days, time format, and per-window timezones |
| backend/internal/models/settings.go | 5/5 | Adds UpdateScheduleWindow struct with proper JSON tags for Days, StartTime, EndTime, and Timezone fields |
| backend/resources/migrations/postgres/022_add_update_schedule.up.sql | 5/5 | Adds scheduling schema by inserting settings, dropping autoUpdateInterval, making projects.auto_update nullable, and adding schedule override columns |

</details>

</details>


kmendell:
I glanced at this and im not sure if i 100% agree with the changes, Image polling shouldnt be project specific:

1. It completely ignore standalone containers then
2. Its inefficent as projects can use teh same images, potentially so hence hwy i have the gloabl image update service with the cache already.
3. The polling should be at the image level, and associated with containers, updating "projects" is not recomended and doesnt work, it should update teh container using the image and just restart the conatinber that how watchtower worked as well. Which i mirroed this logic off of when i was bulding it 
4. The maintenance window is useful but i belive theres a much smplier way to add this , that can be per project or gloabl just based on a simple setting varible for the project or the environment. 

I havent fully looked at it but long story short most of these chnages could be/should be done with the existing service for image updates, while the logic i had was not perfect ill admit that, this seems to have some flaws that i dont think will include everything, this also could technically be a breaking chnage since it removed teh old auto update logic almost entirley.

cabaucom376:
@greptileai

cabaucom376:
@greptileai

frontend/src/lib/layouts/tabbed-page-layout.svelte
```diff
@@ -49,7 +49,7 @@
 <div class={cn('bg-background flex min-h-0 flex-col', className)}>
 	<Tabs.Root value={selectedTab} class="flex min-h-0 w-full flex-1 flex-col">
 		<div
-			class="bg-background/95 sticky top-0 border-b backdrop-blur transition-all duration-300"
+			class="bg-background/95 sticky top-0 z-20 border-b backdrop-blur transition-all duration-300"
```
kmendell: Adding a z-index breaks the layout 

Image

kmendell: In general we should leave the z-index alone, most of the components already are correctly layered.

frontend/src/lib/services/project-service.ts
```diff
@@ -9,7 +9,23 @@ export class ProjectService extends BaseAPIService {
 		const envId = await environmentStore.getCurrentEnvironmentId();
 		const params = transformPaginationParams(options);
 		const res = await this.api.get(`/environments/${envId}/projects`, { params });
-		return res.data;
+		const data = res.data;
+
+		if (data.data && Array.isArray(data.data)) {
```
kmendell: This logic is overly complex and should be able to be simplified if the correct data is being returned it should be included in the res.data.

frontend/src/lib/services/project-service.ts
```diff
@@ -38,7 +54,18 @@ export class ProjectService extends BaseAPIService {
 			this.api.get(`/environments/${envId}/projects/${projectId}`)
 		);
 
-		return response.project ? response.project : (response as Project);
+		const project = response.project ? response.project : (response as Project);
```
kmendell, line 43: Same here

frontend/src/routes/projects/components/ProjectSettings.svelte
kmendell, line 1: Im not a fan of this UI component, its very bulky, i think for the date picker and time picker we can use somethign like this https://shadcn-svelte.com/blocks/calendar#calendar-17, In addition im also not a fan of the "Overide" wording and orange box that goes around it

kmendell: This is the case for the settings page as well

frontend/src/routes/onboarding/docker/+page.svelte
```diff
@@ -57,10 +59,22 @@
...
 		pollingInterval: z.number().int().min(5).max(10080),
 		autoUpdate: z.boolean(),
-		autoUpdateInterval: z.number().int(),
-		dockerPruneMode: z.enum(['all', 'dangling'])
+		dockerPruneMode: z.enum(['all', 'dangling']),
+		updateScheduleEnabled: z.boolean(),
+		updateScheduleWindows: z.array(
+			z.object({
+				days: z.array(z.enum(['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday'])),
+				startTime: z.string().regex(/^([0-1]?[0-9]|2[0-3]):[0-5][0-9]$/, 'Invalid time format'),
```
kmendell: zod/v4 has a date parsing schema https://zod.dev/api#dates

frontend/src/routes/projects/components/ProjectSettings.svelte
```diff
@@ -0,0 +1,196 @@
...
+		updateScheduleEnabled: project.updateScheduleEnabled ?? globalSettings?.updateScheduleEnabled ?? false,
+		updateScheduleWindows: project.updateScheduleWindows ?? globalSettings?.updateScheduleWindows ?? []
+	});
+
+	const formSchema = z.object({
+		autoUpdate: z.boolean(),
+		updateScheduleEnabled: z.boolean(),
+		updateScheduleWindows: z.array(
+			z.object({
+				days: z.array(z.enum(['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday'])),
```
kmendell: zod/v4 has a date parsing schema https://zod.dev/api#dates

kmendell: In additional most logic like this should be kept to the backend, the main purpose of the frontned is just static files, (this is something i know i need to fix other places just its a bigger task)

frontend/src/routes/settings/docker/+page.svelte
```diff
@@ -31,7 +32,15 @@
 		pollingEnabled: z.boolean(),
 		pollingInterval: z.number().int().min(5).max(10080),
 		autoUpdate: z.boolean(),
-		autoUpdateInterval: z.number().int(),
+		updateScheduleEnabled: z.boolean(),
+		updateScheduleWindows: z.array(
+			z.object({
+				days: z.array(z.enum(['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday'])),
```
kmendell: Same thing here about date validation

backend/internal/services/project_service.go
```diff
@@ -920,3 +926,86 @@ func (s *ProjectService) calculateProjectStatus(services []ProjectServiceInfo) m
 	}
 	return models.ProjectStatusUnknown
 }
+
+func (s *ProjectService) UpdateProjectSettings(ctx context.Context, projectID string, updates dto.UpdateProjectSettingsDto) error {
```
kmendell: There are convience functions for getting and setting settings data, The settings use reflect. See the settibgs_service ListSettings and SetSettings fucntions , and the ToVariablesSlice fucntions, this will esnure consistenit behavior without a lot of this unneeded logic.

frontend/src/lib/layouts/tabbed-page-layout.svelte
cabaucom376: There was a purpose before as the content of the layout overflowed the sticky header whenever there wasn't a floating header but I cant seem to find where I saw that before so I guess its good to remove for now, just keep an eye out for that scenario.

cabaucom376:
@kmendell I'm starting to think it's best to just scrap this and reimplement it better

kmendell:
@cabaucom376 Up to you, i can deal with the merge conflicts if thats your concern

cabaucom376:
I honestly didn't know that the `com.ofkm.arcane.updater=false` label feature existed until yesterday. What if we just expanded the control over that label to support scheduling and performed some sort of deploy-time injection using project-level frontend settings for convenience, so that would apply to the containers within the project.

cabaucom376:
@kmendell I am thinking of:
1. Removing the project level settings
2. Providing some new labels for managing container level auto update scheduling
3. Allowing labels to specify `com.ofkm.arcane.updater=true` with global disabled for explicit opt-in
4. Adapt the project level setting ui to instead just apply labels to the services within it?

To me this strikes a nice balance between not needing to do/store anything project level but provide an easy way for projects to orchestrate containers together. This would also be portable if you want to copy your setup to another machine you don't have magic database values that won't transfer. Also could be a cool pattern for additional labels down the road for container level things.  Think like how Dockge has container level inputs for editing the compose file in a visual way, but not for every property, unless thats something you would want.

Also as a follow up to the most recent comment on the issue we could consider storing the data as a cron schedule.

PR #835, opened 2025-11-02
feat: command pallette

cabaucom376:
Just an experiment for now


kmendell:
We do not need to make our own component for this. Shadcn-svelet has a command component https://shadcn-svelte.com/docs/components/command. This will alleviate the need for extra code taht is not needed.

That being said, im not sure i see the point of this is this something that people want? i dont want to add bloat to the UI if it wont be a used feature.

kmendell:
closing for now

PR #1010, opened 2025-11-30
fix: Pass project object to StartOptions

### LINKED ISSUE #1005: 🐞 Bug: Docker configs not properly handled in projects (compose)
### Bug Description

In Docker Standalone (not Swarm), docker configs are normally supported when using docker compose as per the compose spec. When deploying a project in Arcane, the configs do not appear to work. Using the identical compose file from the docker command line works as expected.

### Steps To Reproduce

1. Create a new project
2. In the compose content, add a top-level configs: section with at least 1 config, and configure one of the services to attach the config
   ```yaml
   configs:
     some_config:
       content: |
         Some text content
   services:
     some-service:
       configs:
         - source: some_config
           target: /some/path
   ```
3. Save and deploy the project

### Expected Behavior

The specified config file attached to the desired path in the container.

### Actual Behavior

The file is not attached and therefore does not exist in the container.

### Screenshots

_No response_

### Arcane Version

v1 9.0

### Operating System

Linux

### Browser

Chrome 142, others

### Docker Version

28.3.1

### Additional Context

While docker standalone does not support managing configs via the CLI, these have been supported for some time when using compose. Using an identical compose.yaml and bringing it up via the CLI with `docker compose up -d` works as expected on the same docker instance.

chadweimer:
I believe this fixes #1005 by passing the entire project object to start, thus allowing it to create configs and secrets that use either `environment` or `content` (rather than `file`), which follows a separate code path in `github.com/docker/compose`.

**Disclaimer** Greptiles Reviews uses AI, make sure to check over its work

<h2>Greptile Overview</h2>

<details><summary><h3>Greptile Summary</h3></summary>

Passed the entire project object to `api.StartOptions` to properly support Docker configs and secrets using `content` or `environment` fields instead of only file-based configs.

- Fixed Docker configs not being properly attached to containers when using `content` or `environment` fields
- The docker/compose library requires the full project object in `StartOptions` to handle non-file-based configs/secrets
- Previously only `Services` and `Wait` fields were populated in `StartOptions`, which caused the library to skip configs/secrets with inline content

</details>
<details><summary><h3>Confidence Score: 5/5</h3></summary>

- This PR is safe to merge with minimal risk
- The change is a minimal, focused fix that adds a missing field to properly support Docker compose configs and secrets. The project object is already available and being used elsewhere in the same function, so passing it to StartOptions is safe. This aligns with how the docker/compose library expects to receive configuration data.
- No files require special attention
</details>

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| backend/internal/utils/projects/cmds.go | 5/5 | Added `Project` field to `StartOptions` to enable proper handling of Docker configs and secrets with `content` or `environment` fields |

</details>

</details>

### DIFF
--- a/tests/setup/project.data.ts
+++ b/tests/setup/project.data.ts
@@ -1,14 +1,24 @@
-export const TEST_COMPOSE_YAML = `services:
+export const TEST_COMPOSE_YAML = `configs:
+  some_content:
+    content: |
+      This is a test config file.
+      It can contain multiple lines of text.
+      Used for testing purposes.
+
+services:
   redis:
     image: redis:latest
     container_name: \${CONTAINER_NAME}
+    configs:
+      - source: some_content
+        target: /etc/some_content.txt
+    command: /bin/sh -c 'cat /etc/some_content.txt && redis-server'
     ports:
       - "8081:81"
       - "6379:6379"
       - "6378:6378"
     volumes:
       - redis_data:/data
-    restart: unless-stopped
 
 volumes:
   redis_data:
--- a/tests/spec/project.spec.ts
+++ b/tests/spec/project.spec.ts
@@ -210,6 +210,9 @@ test.describe('New Compose Project Page', () => {
 
     const serviceBadge = page.locator('text=redis').first();
     await expect(serviceBadge).toBeVisible();
+
+    const statusBadge = page.locator('text=Running').first();
+    await expect(statusBadge).toBeVisible();
   });
 });
 


kmendell:
Ive never messed with passing stuff like this, do you have a example compose i can try?

chadweimer:
@kmendell Yeah, to reproduce this is what I was trying. It fails since the `cat` command errors due to the file missing.

```yaml
configs:
  some_file:
    content: |
      Some
      Text
services:
  alpine:
    image: alpine
    configs:
      - source: some_file
        target: /tmp/file.txt
    entrypoint: /bin/sh
    command: -c 'cat /tmp/file.txt && sleep infinity'
```

chadweimer:
I'll try to get a new docker image from this and give it a test tomorrow. Will check here to see if you've already done that for me :smile:

kmendell:
Its 2am for me right now lol, but i will try to do some research

chadweimer:
Well, I couldn't sleep without trying it. I got my dev environment working and was able to confirm this does fix this issue. I re-confirmed the problem with latest main branch, then confirmed the configs are properly configured after the fix here. Let me know if there tests you'd want to add for this. I've not yet familiarized myself with that part of the codebase.

kmendell:
@chadweimer For the test, honestly we could just change what project is passed in project.spec.ts for playwright and just pass it custom content or something to make sure its loaded, it may be tricky to 100% confirm though

tests/setup/project.data.ts
```diff
@@ -1,14 +1,24 @@
...
+      - source: some_content
+        target: /etc/some_content.txt
+    command: /bin/sh -c 'cat /etc/some_content.txt && redis-server'
     ports:
       - "8081:81"
       - "6379:6379"
       - "6378:6378"
     volumes:
       - redis_data:/data
-    restart: unless-stopped
```
chadweimer, line 11: Removing this is necessary for the new confirmation of the container being in the running state, which could be incorrectly detected if the container is in a restart loop.

chadweimer:
@kmendell I took a shot at a small tweak to the existing E2E test as you suggested. Let me know what you think. I was initially considering adding a whole separate test, but realized the existing functionality could be preserved with just an adjustment to the test project compose file.

PR #1035, opened 2025-12-02
chore(translations): Extended german translation

NeoCortex3:
Some more work on the german translation.


NeoCortex3:
All these Bot-Checks and automation when submitting a PR is realy confusing...
To be honest, I have no idea what github is doing here.

PR #1075, opened 2025-12-05
chore(translations): My final german translation

NeoCortex3:
Final PR for the german translation.
These are all 1378 lines translated.

There is still some work todo.
For example I think in some places I mixed the formal wording "Sie" and "Du",
also it wasn't clear to me where some of the strings are used, so wording can be off in these places.

But I think it is a good basis and can be easily polished.


kmendell (approved):

kmendell:
Thank you for your contributions :)

NeoCortex3:
> Thank you for your contributions :)

Thanks for the great tool

PR #1278, opened 2025-12-29
chore(translations): update translations via Crowdin

getarcaneappbot:


NeoCortex3:
grafik

This changes "Volumes" back to "Lautstärken" which is more than wrong in german!
A Docker Volume is somewhat like a virtual Harddrive (more or less).
Volume translated 1:1 in german is "Lautstärke" which is the volume of a sound or music.
The context is wrong here!

In the context of Docker we also just call it a volume. If people keep changing this via cowdin (with the help of AI I guess), it's a battle I can't win and I won't fight.

kmendell:
@NeoCortex3 That was probably the pretranslation stuff from crowdin.... I can turn it off if its messing up certain things.

NeoCortex3:
> @NeoCortex3 That was probably the pretranslation stuff from crowdin.... I can turn it off if its messing up certain things.

pretranslation? So these are new Strings? I'm realy not familiar with crowdin.
I thought it changed allready translated strings back. In that case I'm sorry, my fault.

PR #1291, opened 2025-12-30
chore(translations): update german translation

NeoCortex3:
updated translation and fixed some spelling errors.


kmendell:
@NeoCortex3  sorry i had to fix some weirdness in those crowdin prs, but yes pre trnslation only happens on new strings.

PR #1629, opened 2026-02-02
feat: Git Sync Folder Implementation

### LINKED ISSUE #1619: ⚡️ Feature: Git Sync should pull entire project directory, not just a single compose file
### Feature Description

Git Sync currently only pulls the single `compose.yaml` file specified in `dockerComposePath`. This breaks projects that use Docker Compose's `extends` directive (or `include`) to reference additional files in the same repository directory.

**Example structure in the Git repo:**

```
myproject/
├── compose.yml
├── base-services.yml
├── networking.yml
└── .env.example
```

Where `compose.yml` uses:

```yaml
services:
  app:
    extends:
      file: base-services.yml
      service: app-base
    ports:
      - "8080:80"
```

When Git Sync pulls only `compose.yml`, Docker Compose fails because `base-services.yml` is missing from the project directory. The same issue applies to `include:` directives and any relative file references within compose files.

### Problem It Solves

Many real-world Docker Compose setups factor common service definitions into base files to avoid duplication across stacks. The `extends` feature is a core part of the Compose spec and is widely used for:

- Shared base service configs (logging, healthchecks, resource limits)
- Environment-specific overrides (`compose.yml` extends `base.yml`)
- Config files, custom entrypoint scripts, or Dockerfiles referenced via relative paths in `build:` contexts

### Proposed Solution

Instead of syncing only the file at `ComposePath`, Git Sync should pull **the entire directory** containing the compose file (and optionally its subdirectories). The `dockerComposePath` field would still identify the entrypoint compose file, but all sibling files in that directory would be synced alongside it.

For example, given `dockerComposePath: "stacks/myproject/compose.yml"`, Arcane should sync the full `stacks/myproject/` directory into the project folder — not just the single `compose.yml`.

**Alternatively**, a new optional field like `syncDirectory` or `additionalFiles` could give users explicit control:

```json
{
  "syncName": "my-project",
  "gitRepo": "my-repo-id",
  "branch": "main",
  "dockerComposePath": "stacks/myproject/compose.yml",
  "syncDirectory": true,
  "autoSync": true,
  "syncInterval": 5
}
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Would you be willing to work on this feature?

- [ ] Yes, I'd like to implement this feature
- [x] I could help with parts of this feature
- [ ] No, I'm just suggesting the feature

sasha-id:
Fixes: https://github.com/getarcaneapp/arcane/issues/1619

I've implemented the directory sync feature for Git Sync. Here's what was built:
Directory sync capability that allows Git Sync to pull entire directories instead of just single compose files, supporting Docker Compose extends, include, and relative file references.

- Recursive sync: Syncs compose file directory and all subdirectories
- Safety limits: 500 files max, 50MB total, skip binaries >10MB
- Default enabled: syncDirectory=true for all syncs (backward compatible)
- File tracking: Synced files stored in DB for cleanup on updates

This PR implements directory sync for Git Sync, allowing Arcane to pull an entire compose-file directory (including subdirectories) rather than just a single compose file. This enables support for Docker Compose `extends`, `include`, and relative file references that span multiple files. The feature is backward-compatible (`syncDirectory=true` is the new default), adds configurable safety limits (file count, total size, binary-file size) enforced at both the per-sync and environment levels, and tracks synced file paths in the database for accurate cleanup on subsequent syncs.

Key changes:
- **`WalkDirectory`** in `backend/pkg/gitutil/git.go`: recursive directory traversal with binary detection via `net/http.DetectContentType` and null-byte fallback.
- **`WriteSyncedDirectory` / `CleanupRemovedFiles`** in `backend/pkg/projects/fs_writer.go`: secure multi-file writes with `IsSafeSubdirectory` path-traversal guards and best-effort empty-directory cleanup.
- **`GitOpsSyncService`**: two-tier limit system (per-sync values capped by environment-wide settings) via `effectiveIntLimit`.
- **`enrichWithDirectoryFiles`** in `project_service.go`: displays non-compose, non-env files from the project directory in the UI using `os.OpenRoot` for sandboxed access.
- **Migrations** (Postgres + SQLite): adds `sync_directory`, `synced_files`, and three `max_*` columns with correct defaults.
- **Frontend**: new "Sync Files" toggle in the GitOps sync dialog, read-only directory-file viewer in the project page tree, and environment-level limit controls in GeneralTab.

Issues found:
- `appendSyncFile` in `git.go` reads the full file content into memory before applying the binary-size skip — large binary files cause a transient memory spike before being discarded.
- `effectiveIntLimit`/`effectiveInt64Limit` treat a sync-level `0` as "defer to environment cap", but all API documentation and model comments state "0 means unlimited" — this semantic mismatch can surprise API consumers who set `maxSyncFiles=0` expecting to remove the per-sync cap.
- The compose-file lookup in `walkAndParseSyncDirectory` relies on an implicit invariant (the walk root is always the direct parent of the compose file) that should be documented to prevent future regressions.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

Core logic is well-structured and tested, but two concrete issues — unexpected memory pressure on binary-heavy repos and a doc/behaviour mismatch that silently caps 'unlimited' syncs — should be resolved before merging.

The implementation is feature-complete with solid path-traversal guards, proper migration files for both databases, and good test coverage. However, the memory issue in appendSyncFile is a real reliability concern for users with large binaries in their repos, and the '0 means unlimited' semantic contradiction across multiple API types and model comments is a correctness/trust issue that will be hard to fix later without a breaking change.

backend/pkg/gitutil/git.go (appendSyncFile memory), backend/internal/services/gitops_sync_service.go and types/gitops/gitops.go (effectiveIntLimit semantics + documentation)
</details>

| Filename | Overview |
|----------|----------|
| backend/pkg/gitutil/git.go | Adds WalkDirectory + helper functions for recursive repo traversal; binary detection via net/http is solid, but full file read before binary-size skip wastes memory for large repos. |
| backend/internal/services/gitops_sync_service.go | Core directory-sync orchestration; limit-enforcement logic is solid but "0 = unlimited" semantics in effectiveIntLimit contradict API documentation, and compose-file lookup relies on an undocumented invariant. |
| backend/pkg/projects/fs_writer.go | Adds WriteSyncedDirectory and CleanupRemovedFiles with proper path-traversal guards (IsSafeSubdirectory) and best-effort cleanup logic; well structured. |
| backend/internal/services/project_service.go | enrichWithDirectoryFiles uses os.OpenRoot for sandboxed access, skips symlinks, limits files to 1 MB, and detects binary via null-byte check — good security hygiene. |
| frontend/src/routes/(app)/projects/[projectId]/+page.svelte | Adds directoryFiles tree rendering in both tablet and desktop layouts; dir: prefix scheme for selectedFile is straightforward, and all directory files are shown as read-only. |
| types/gitops/gitops.go | New fields are well-documented but the "0 means unlimited" comment is misleading given that effectiveIntLimit defers to the environment cap when sync value is 0. |
| backend/internal/models/gitops_sync.go | Model expanded with SyncDirectory, SyncedFiles, and three max-* limit columns; GORM tags and defaults align with migration SQL. |
| backend/internal/services/settings_service.go | Three new Git Sync limit defaults added consistently with other setting variables; defaults match the constants in gitops_sync_service.go. |
| backend/resources/migrations/postgres/041_add_directory_sync.up.sql | Adds sync_directory, synced_files, and three max_* columns with correct defaults; index on sync_directory is appropriate. |
| backend/resources/migrations/sqlite/041_add_directory_sync.down.sql | SQLite down migration reconstructs the table to drop columns; approach is correct for SQLite's lack of DROP COLUMN support in older versions. |
| backend/internal/services/gitops_sync_service_test.go | Good tests covering environment limit inheritance, tighter sync limits, and the zero-means-defer-to-env case. |
| frontend/src/routes/(app)/environments/[id]/components/GeneralTab.svelte | New Git Sync File Limits section renders three numeric inputs with correct i18n keys and error bindings. |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant S as GitOpsSyncService
    participant G as gitutil.Client
    participant FS as fs_writer
    participant DB as Database

    S->>S: getEffectiveSyncLimits(ctx, sync)
    note over S: min(sync limits, env limits)

    S->>G: WalkDirectory(repoPath, composePath, maxFiles, maxTotalSize, maxBinarySize)
    loop fs.WalkDir
        G->>G: walkSyncEntry(path, entry)
        G->>G: appendSyncFile(path)
        note over G: ReadFile → detect binary → check size limits
    end
    G-->>S: DirectoryWalkResult{Files, TotalFiles, TotalSize, SkippedBinaries}

    S->>S: find compose file in walkResult (by base name)
    S->>S: getOrCreateProjectInternal(composeContent, nil)

    S->>FS: CleanupRemovedFiles(projectsRoot, projectPath, oldFiles, newFiles)
    FS->>FS: delete removed files, cleanupEmptyDirs

    S->>FS: WriteSyncedDirectory(projectsRoot, projectPath, syncFiles)
    FS->>FS: validate paths (IsSafeSubdirectory)
    FS-->>S: writtenPaths

    S->>DB: updateSyncStatusWithFiles(id, success, commitHash, syncedFiles)
    DB-->>S: ok
```
</details>

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `frontend/src/routes/(app)/projects/[projectId]/+page.svelte`, line 715-721 ([link](https://github.com/getarcaneapp/arcane/blob/64e9d149e3ea0967b5f6014d1d53e6139db27411/frontend/src/routes/(app)/projects/[projectId]/+page.svelte#L715-L721)) 

   <a href="#">P1</a> **Validation and diff bindings stripped from tablet compose panel**

   The tablet layout's compose `CodePanel` lost `bind:hasErrors`, `bind:validationReady`, `fileId`, `originalValue`, `enableDiff`, and `editorContext`. Without `bind:hasErrors={composeHasErrors}` and `bind:validationReady={composeValidationReady}`, the save-button guard that prevents submitting invalid YAML will stop working on tablet viewports — users could persist broken compose files without seeing any validation error. The diff view is also silently disabled.

   These same bindings are correctly kept in the non-tablet code path (around line 857). They should be restored in the tablet branch as well.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: frontend/src/routes/(app)/projects/[projectId]/+page.svelte
   Line: 715-721

   Comment:
   **Validation and diff bindings stripped from tablet compose panel**

   The tablet layout's compose `CodePanel` lost `bind:hasErrors`, `bind:validationReady`, `fileId`, `originalValue`, `enableDiff`, and `editorContext`. Without `bind:hasErrors={composeHasErrors}` and `bind:validationReady={composeValidationReady}`, the save-button guard that prevents submitting invalid YAML will stop working on tablet viewports — users could persist broken compose files without seeing any validation error. The diff view is also silently disabled.

   These same bindings are correctly kept in the non-tablet code path (around line 857). They should be restored in the tablet branch as well.

   `````
   </details>

2. `frontend/src/routes/(app)/projects/[projectId]/+page.svelte`, line 734-743 ([link](https://github.com/getarcaneapp/arcane/blob/64e9d149e3ea0967b5f6014d1d53e6139db27411/frontend/src/routes/(app)/projects/[projectId]/+page.svelte#L734-L743)) 

   <a href="#">P2</a> **Hardcoded `language="yaml"` for all directory files**

   The directory sync feature can pull any file type from the repository (shell scripts, JSON, Makefile, `.env`, Nginx configs, etc.). Using `language="yaml"` for every directory file will produce incorrect syntax highlighting for non-YAML files.

   Consider deriving the language from the file extension:

   ```svelte
   {@const dirLang = dirFile.relativePath.endsWith('.json') ? 'json' :
      dirFile.relativePath.endsWith('.sh') ? 'shell' :
      dirFile.relativePath.endsWith('.env') ? 'env' : 'yaml'}
   <CodePanel
       open={true}
       title={dirFile.relativePath}
       language={dirLang}
       value={dirFile.content}
       readOnly={true}
   />
   ```

   This pattern is repeated in two places (tablet and desktop views).

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: frontend/src/routes/(app)/projects/[projectId]/+page.svelte
   Line: 734-743

   Comment:
   **Hardcoded `language="yaml"` for all directory files**

   The directory sync feature can pull any file type from the repository (shell scripts, JSON, Makefile, `.env`, Nginx configs, etc.). Using `language="yaml"` for every directory file will produce incorrect syntax highlighting for non-YAML files.

   Consider deriving the language from the file extension:

   ```svelte
   {@const dirLang = dirFile.relativePath.endsWith('.json') ? 'json' :
      dirFile.relativePath.endsWith('.sh') ? 'shell' :
      dirFile.relativePath.endsWith('.env') ? 'env' : 'yaml'}
   <CodePanel
       open={true}
       title={dirFile.relativePath}
       language={dirLang}
       value={dirFile.content}
       readOnly={true}
   />
   ```

   This pattern is repeated in two places (tablet and desktop views).

   `````
   </details>
</details>

`````markdown
Path: backend/pkg/gitutil/git.go
Line: 612-621

Comment:
**Large binary files fully read before size check**

`root.ReadFile(path)` loads the complete file into memory before the binary-size guard on line 620 runs. A 100 MB image that will ultimately be skipped because it exceeds `maxBinarySize` still gets entirely buffered, causing a transient memory spike per such file.

The fix is to stat the file first using the `fs.DirEntry` available in `walkSyncEntry` and skip early if the file is already known to be larger than `maxBinarySize`:

```go
// In walkSyncEntry, before calling appendSyncFile:
if limits.maxBinarySize > 0 {
    if info, err := d.Info(); err == nil && info.Size() > limits.maxBinarySize {
        result.SkippedBinaries++
        return nil
    }
}
```

Alternatively, read only the first 512 bytes for binary detection and then use `d.Info().Size()` for the size check, avoiding a full read of large files that will be discarded.

---

Path: backend/internal/services/gitops_sync_service.go
Line: 66-88

Comment:
**"0 means unlimited" documentation contradicts actual behaviour**

The `GitOpsSync` model, `CreateSyncRequest`, `UpdateSyncRequest`, and `types/gitops/gitops.go` all document `MaxSyncFiles`, `MaxSyncTotalSize`, and `MaxSyncBinarySize` with "0 means unlimited". But `effectiveIntLimit`/`effectiveInt64Limit` treat a sync-level 0 as "defer to the environment cap" — not as truly unlimited:

```go
case syncLimit == 0:
    return environmentLimit   // returns env cap, *not* 0
```

A caller who reads the API docs and sets `maxSyncFiles: 0` to remove the per-sync cap will silently still be bounded by the environment setting. True "unlimited" only happens when both the sync value **and** the environment setting are 0.

This mismatch exists in every file that documents these fields:
- `types/gitops/gitops.go` lines 128, 133, 138, 337, 343, 349, 400, 405, 410, 666
- `backend/internal/models/gitops_sync.go` (the `// 0 = unlimited` inline comments)

The documentation should clarify that 0 at the sync level means "inherit the environment cap", and that a sync cannot opt out of the environment limit. Alternatively, the semantics should be inverted so that a sync-level 0 truly bypasses the environment cap (which may be the more natural expectation for an API user).

---

Path: backend/internal/services/gitops_sync_service.go
Line: 929-930

Comment:
**Compose file lookup relies on an undocumented invariant**

`composeFileName := filepath.Base(sync.ComposePath)` strips the directory prefix entirely. This works correctly today because `WalkDirectory` always roots the walk at the *direct parent* of the compose file — so the compose file appears at the root of the walked tree with just its base name as its relative path.

However, this invariant is nowhere documented. A future refactor that walks a wider directory (e.g., the repo root instead of the compose file's parent) would cause `f.RelativePath == composeFileName` to silently match a wrong file in a subdirectory, or miss the compose file entirely.

Add a comment explaining why the base name comparison is sufficient:

```go
// composeFileName is the base name of the compose file. WalkDirectory roots
// the walk at filepath.Dir(composePath), so the compose file always appears
// at the top level of the walk with RelativePath == filepath.Base(composePath).
composeFileName := filepath.Base(sync.ComposePath)
```

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["fixes"](https://github.com/getarcaneapp/arcane/commit/8796c54c4f674846ce73c9305791f0f7744cbc02) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=21235736)</sub>

> Greptile also left **3 inline comments** on this PR.

**Context used:**

- Rule used - # Golang Pro

Senior Go developer with deep expert... ([source](https://app.greptile.com/review/custom-context?memory=214b40a8-9695-4738-986d-5949b5d65ff1))
- Rule used - # Go Development Patterns

**What:** Code should p... ([source](https://app.greptile.com/review/custom-context?memory=c1082b6a-5fdc-4db8-8419-8a71ccd57636))

### CODE REVIEW

File: backend/internal/services/gitops_sync_service.go [hunk 1]
```diff
@@ -2,36 +2,48 @@ package services
...
+	eventService    *EventService
+	settingsService *SettingsService
 }
 
 const defaultGitSyncTimeout = 5 * time.Minute
 
-func NewGitOpsSyncService(db *database.DB, repoService *GitRepositoryService, projectService *ProjectService, eventService *EventService) *GitOpsSyncService {
+// Directory sync limits
+const (
+	maxSyncFiles      = 500                    // Maximum number of files to sync
```
kmendell: I would think we should make these configurable? Limits are a good idea, but some people could care less about limits and just want everything no matter what.   
  
I think We can add these columns to each Git sync configuration and have these be the defaults.

kmendell:
@sasha-id As anoteghr PR was merged to support .env files, do you wnat to restructure yours to allow this stuff including with the existsing support?

sasha-id:
> @sasha-id As anoteghr PR was merged to support .env files, do you wnat to restructure yours to allow this stuff including with the existsing support?

Go isn't really my strongest skill — I was happy to make the initial contribution, but I think it would be better if someone more experienced with Go took this on to make sure it's done properly.

kmendell:
> > @sasha-id As anoteghr PR was merged to support .env files, do you wnat to restructure yours to allow this stuff including with the existsing support?
> 
> Go isn't really my strongest skill — I was happy to make the initial contribution, but I think it would be better if someone more experienced with Go took this on to make sure it's done properly. 

Understood, Ill try to implement this based off your initial implementation. Thank you :)

sasha-id:
Conflict resolved + few improvements, please review and merge

kmendell:
Ill fix the rest of it tomorrow. Thanks for updating it :)

kmendell (approved):

kmendell:
Thanks!

lordraiden:
Is this already deployed? I dont see the way to configure it

PR #1875, opened 2026-02-24
feat(autoupdate/compose): implement docker compose aware update logic

spupuz:
image

# Container Update Flow: Logic and Scenarios

Arcane's new update architecture explicitly divides the update logic based on the nature of the container (Standalone vs Docker Compose) and acts intelligently to minimize downtime.

There are three main actors in this logic:
1. **User Setting**: The `useComposeUpdate` flag in `Local Docker -> Settings`.
2. **Manual Action**: The "Update" button on individual containers or "Redeploy/Update" on projects.
3. **Auto-Updater**: The background system job (scheduled for automatic execution).

Here is how the code behaves in various scenarios:

## Scenario 1: Standalone Container (Non-Compose)
This is the traditional behavior, unaltered by the new logic.
- **Manual Trigger / Auto-Update**: The system recognizes that the container is not part of a project (absence of the `com.docker.compose.project` label).
- **Executed Flow**: 
  1. `docker stop [id]`
  2. `docker rm -f [id]`
  3. `docker run ...` (Recreates the container parsing the original configuration).

## Scenario 2: Container belonging to a Compose Project
In this scenario, Arcane enters the `ComposeAware` branch. The behavior depends on the global setting configured by the user.

#### If the Update is invoked by the "Auto-Update" Job (Background)
Arcane must prevent downtime and restart **exclusively** what is strictly necessary, and only if the remote image has a truly divergent hash.
- **Executed Flow**:
  1. The DB signals that one or more project containers have a pending update.
  2. Arcane groups all containers marked for update based on their project name (avoiding duplicate pulls).
  3. For each group, it calls `UpdateProjectServices`:
     - The exact *Service Name* (e.g., `web`, `db`) is isolated from the affected container.
     - Executes `docker compose pull [service_names...]` (updates the image only for the services needing the update).
     - Executes a trick: `docker stop && docker rm` (or `docker compose stop`) on the single service to force its recreation, ensuring partial downtime only for the updating service.
     - Executes `docker compose up -d [service_names...]` to bring it back up, automatically reattaching it to networks/volumes and dependencies.

#### If the Update is manually invoked by the "Redeploy/Update" button in the "Projects" UI
Arcane uses a *brute force* approach, since it expects the user to know what they are doing or they are explicitly forcing a reload of the entire project configuration, regardless of the Image Poller.
- **Executed Flow**:
  1. No preventive hash checks (it doesn't matter if the image is already up-to-date, it skips nothing).
  2. Enters `UpdateProjectServices` *without service limitations*.
  3. Executes `docker compose pull` (downloads layers if there are new ones online for any service).
  4. Executes `docker compose up -d` (Docker recreates **only** the containers whose configuration or image has mutated. The intact ones will remain Up without downtime).

## Resiliency of Auto-Update (Hash Checks)
The Auto-Update system prevents unnecessary downtime by implementing double security checks:
1. Before proceeding with *any* mechanics described above, the verification chain is instantiated via the `ApplyPending` method.
2. Arcane extracts the *Local Digest SHA256* of the image currently bound to the containers associated with "Pending" records.
3. Requests the *Remote Digest SHA256* of the target image on the fly from the Internet.
4. **If the Digest matches (Local == Remote)**: the routine is bypassed. A log `"status":"skipped", "error":"image already up to date"` is registered, and the container remains alive and unchanged. The green checkmark ✅ appears in the Updates column.
5. **If the Digest does not match**: the Update pipeline (Standalone or ComposeAware) is officially authorized, and the container is updated.

This PR introduces a compose-aware auto-update architecture that intelligently routes container updates through either `docker compose` commands (for containers belonging to tracked Compose projects) or the existing standalone stop/rm/run path. It addresses several previously-flagged concerns — service deduplication via `projectToSeenServices`, the shallow-copy issue in `ComposePull` (replaced with `WithSelectedServices`), the shell-injection risk in global teardown (replaced with `execFileSync`), and the Svelte 5 `$effect`→`$derived` anti-pattern in `JobsTab`.

Key additions:
- `GetProjectByComposeName` with a two-level fallback (exact DB → in-memory cache → full cache rebuild) in `project_service.go`
- `UpdateProjectServices` orchestrating pull → stop → up for selected services
- Pre-scan + lazy-registration pattern in `restartContainersUsingOldIDs` to group compose containers before triggering project-level updates
- `ComposePull` / `ComposeStop` helpers with correct empty-slice guards
- `data-action` attribute on `ArcaneButton` enabling stable test selectors

Two logic issues remain:

- **Double-update on lazy registration** (`updater_service.go:1455`): when a container's initial inspect fails in the pre-scan but succeeds lazily in the main loop, `shouldUpdateComposeProjectInternal`'s count-based re-trigger fires `UpdateProjectServices` a second time, re-stopping and re-starting services that were already updated in the first call — producing unnecessary extra downtime.
- **Stale cache on project rename** (`project_service.go:259`): `lookupProjectByCachedComposeName` returns a project fetched by cached ID without verifying its normalized name still matches the compose label, so a renamed project can be returned (and updated) indefinitely for the wrong compose project name.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

Not safe to merge yet — two P1 logic bugs can cause unintended extra downtime and incorrect project routing in specific but realistic scenarios.

The PR successfully addresses the majority of previously flagged issues (deduplication, shallow copy, shell injection, $effect anti-pattern). However, two new P1 bugs were found: (1) the count-based re-trigger in `shouldUpdateComposeProjectInternal` can double-stop/restart already-updated services when lazy registration adds new services, and (2) `lookupProjectByCachedComposeName` can permanently return a renamed project for the wrong compose label. Both bugs affect the correctness of the auto-update path and require targeted fixes before merge.

`backend/internal/services/updater_service.go` (double-update logic near line 1455) and `backend/internal/services/project_service.go` (stale cache validation near line 259) require attention.
</details>

| Filename | Overview |
|----------|----------|
| backend/internal/services/updater_service.go | Core of the compose-aware update logic: pre-scan groups services by project, lazy registration handles inspect failures, but `shouldUpdateComposeProjectInternal`'s count-based re-trigger can double-update already-processed services. |
| backend/internal/services/project_service.go | Adds `GetProjectByComposeName` with an in-memory cache and fallback rebuild; stale-on-rename bug in cache validation; several unexported methods missing the required "Internal" suffix. |
| backend/pkg/projects/cmds.go | Adds `ComposePull`, `ComposeStop`, and `filterProjectServicesForPullInternal`; `ComposeStop` guards against empty-slice (stops all) correctly; uses library's `WithSelectedServices` instead of the previously flagged shallow copy. |
| backend/internal/services/project_service_test.go | New tests cover exact match, normalized fallback, display-name DB path, and stale-cache invalidation on deletion for `GetProjectByComposeName`; build-path translation assertions updated. |
| backend/internal/services/updater_service_test.go | Adds table tests for `lookupComposeProjectIDInternal`, `lazyRegisterComposeProjectInternal`, and `shouldUpdateComposeProjectInternal`; coverage is adequate for the helper functions. |
| frontend/src/routes/(app)/environments/[id]/components/JobsTab.svelte | Replaces `$effect`-based mutable `SvelteSet` mutation with `$derived.by` — correct idiomatic Svelte 5 pattern; toggle logic properly creates a new set before updating the form input. |
| tests/setup/global-teardown.ts | Switches from shell-interpolation `execSync` to argument-array `execFileSync` for Docker removal of root-owned directories — eliminates the previously flagged shell injection vector. |

</details>

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[Auto-Update: ApplyPending] --> B[restartContainersUsingOldIDs]
    B --> C[Pre-scan sorted containers]
    C --> D{inspect OK?}
    D -- Yes --> E[Add serviceName to projectToServices map]
    D -- No --> F[Skip — inspect is nil]
    E --> G[Main loop: iterate sorted containers]
    F --> G
    G --> H{lazy re-inspect needed?}
    H -- Yes --> I[lazyRegisterComposeProjectInternal\nadds service to projectToServices]
    H -- No --> J{compose project found\nin projectNameToID?}
    I --> J
    J -- Yes --> K{shouldUpdateComposeProject?\nserviceCount > lastCount}
    J -- No --> L[Standalone update\nstop / rm / run]
    K -- Yes --> M[UpdateProjectServices\nComposePull → ComposeStop → ComposeUp]
    K -- No --> N[Mark as updated\nproject already processed]
    M --> O{ComposeUp ok?}
    O -- Yes --> P[Set project status Running\nLog event]
    O -- No --> Q[Set project status Stopped\nReturn error]
```
</details>

`````markdown
Path: backend/internal/services/updater_service.go
Line: 1455-1463

Comment:
**Double-update of already-processed services via count-based re-trigger**

`shouldUpdateComposeProjectInternal` is intended to fire a second `UpdateProjectServices` call when lazy registration discovers more services than the initial pre-scan. However, this also re-applies the update to services that were already processed in the first call, causing unnecessary extra downtime.

**Concrete trace** — project `P` with services `A` (inspect OK in pre-scan) and `B` (inspect fails in pre-scan):

1. Pre-scan: `projectToServices[P.ID] = ["A"]`
2. Main loop, container for `A`: `svcs = ["A"]`, `shouldUpdate(_, P, 1)` → `true` → `UpdateProjectServices(P, ["A"])` → service A is stopped and restarted. `updatedProjectServiceCounts[P.ID] = 1`.
3. Main loop, container for `B` (lazy inspect succeeds): `lazyRegisterComposeProjectInternal` appends `B` → `projectToServices[P.ID] = ["A", "B"]`. `svcs = ["A", "B"]`, `shouldUpdate(_, P, 2)` → `2 > 1` → `true` → `UpdateProjectServices(P, ["A", "B"])` fires again.

Service `A` is now stopped and restarted twice — contradicting the PR's goal of minimising downtime. One approach is to only run `UpdateProjectServices` for the *newly added* services in the re-trigger (`["B"]` only), rather than the full accumulated slice.

---

Path: backend/internal/services/project_service.go
Line: 259-271

Comment:
**Stale cache entry returned after project rename**

`lookupProjectByCachedComposeName` fetches a project by its cached ID but never verifies that the returned project's normalized name still matches `normalizedName`. This means a project rename can produce an incorrect result indefinitely.

**Scenario:** Project `"My App!"` (normalized: `"myapp"`) is stored in the cache as `"myapp" → "proj-id-1"`. The project is later renamed to `"New Service"` (normalized: `"newservice"`) in Arcane. When a compose container labelled `com.docker.compose.project=myapp` triggers a lookup:

1. Direct DB query (`WHERE name = 'myapp' OR name = 'myapp'`) finds no match — project is now `"New Service"`.
2. Cache hit: `projectID = "proj-id-1"`. DB fetch by ID succeeds (the project still exists).
3. The **renamed** project `"New Service"` is returned for the `"myapp"` compose label — `UpdateProjectServices` is called on the wrong project's directory.

Cache invalidation only fires on `ErrRecordNotFound` (deletion), so the stale entry persists across many auto-update cycles.

Adding a name-match assertion before returning would close the gap:
```go
if normalizeComposeProjectName(project.Name) != normalizedName {
    s.invalidateCachedComposeProjectID(normalizedName)
    return nil, false, nil
}
```

---

Path: backend/internal/services/project_service.go
Line: 212-274

Comment:
**Missing "Internal" suffix on unexported methods**

The project's naming convention requires all unexported functions/methods to carry an `Internal` suffix. The following new methods are missing it:

- `getCachedComposeProjectID` (line 212) → `getCachedComposeProjectIDInternal`
- `cacheComposeProjectID` (line 228) → `cacheComposeProjectIDInternal`
- `invalidateCachedComposeProjectID` (line 242) → `invalidateCachedComposeProjectIDInternal`
- `lookupProjectByCachedComposeName` (line 253) → `lookupProjectByCachedComposeNameInternal`
- `rebuildComposeNameCache` (line 274) → `rebuildComposeNameCacheInternal`

For reference, newly-added functions that correctly follow the convention: `translateBuildPathInternal`, `shouldUpdateComposeProjectInternal`, `lazyRegisterComposeProjectInternal`, `lookupComposeProjectIDInternal`.

**Rule Used:** What: All unexported functions must have the "Inte... ([source](https://app.greptile.com/review/custom-context?memory=306fc233-4d2f-4ac4-bdf7-8059588e8a43))

---

Path: backend/internal/services/project_service.go
Line: 265-266

Comment:
**Context error not wrapped with `%w`**

The original context error is discarded here, breaking the error chain and preventing callers from using `errors.Is(err, context.Canceled)`.

```suggestion
		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
			return nil, false, fmt.Errorf("request canceled or timed out: %w", err)
```

**Rule Used:** # Go Development Patterns

**What:** Code should p... ([source](https://app.greptile.com/review/custom-context?memory=c1082b6a-5fdc-4db8-8419-8a71ccd57636))

`````

</details>

<sub>Reviews (9): Last reviewed commit: ["more"](https://github.com/getarcaneapp/arcane/commit/38f34679d640d20bfd03581319ccc33e00b3dbbe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=23063427)</sub>

> Greptile also left **2 inline comments** on this PR.

**Context used:**

- Rule used - What: All unexported functions must have the "Inte... ([source](https://app.greptile.com/review/custom-context?memory=306fc233-4d2f-4ac4-bdf7-8059588e8a43))
- Rule used - # Go Development Patterns

**What:** Code should p... ([source](https://app.greptile.com/review/custom-context?memory=c1082b6a-5fdc-4db8-8419-8a71ccd57636))

### DIFF
--- a/backend/internal/services/project_service.go
+++ b/backend/internal/services/project_service.go
@@ -209,7 +209,7 @@ func normalizeComposeProjectName(name string) string {
 	return normalized
 }
 
-func (s *ProjectService) getCachedComposeProjectID(normalizedName string) (string, bool) {
+func (s *ProjectService) getCachedComposeProjectIDInternal(normalizedName string) (string, bool) {
 	if normalizedName == "" {
 		return "", false
 	}
@@ -225,7 +225,7 @@ func (s *ProjectService) getCachedComposeProjectID(normalizedName string) (strin
 	return projectID, ok
 }
 
-func (s *ProjectService) cacheComposeProjectID(normalizedName, projectID string) {
+func (s *ProjectService) cacheComposeProjectIDInternal(normalizedName, projectID string) {
 	if normalizedName == "" || projectID == "" {
 		return
 	}
@@ -239,7 +239,7 @@ func (s *ProjectService) cacheComposeProjectID(normalizedName, projectID string)
 	s.composeNameToProjID[normalizedName] = projectID
 }
 
-func (s *ProjectService) invalidateCachedComposeProjectID(normalizedName string) {
+func (s *ProjectService) invalidateCachedComposeProjectIDInternal(normalizedName string) {
 	if normalizedName == "" {
 		return
 	}
@@ -250,8 +250,8 @@ func (s *ProjectService) invalidateCachedComposeProjectID(normalizedName string)
 	delete(s.composeNameToProjID, normalizedName)
 }
 
-func (s *ProjectService) lookupProjectByCachedComposeName(ctx context.Context, normalizedName string) (*models.Project, bool, error) {
-	projectID, ok := s.getCachedComposeProjectID(normalizedName)
+func (s *ProjectService) lookupProjectByCachedComposeNameInternal(ctx context.Context, normalizedName string) (*models.Project, bool, error) {
+	projectID, ok := s.getCachedComposeProjectIDInternal(normalizedName)
 	if !ok {
 		return nil, false, nil
 	}
@@ -259,19 +259,23 @@ func (s *ProjectService) lookupProjectByCachedComposeName(ctx context.Context, n
 	var project models.Project
 	if err := s.db.WithContext(ctx).Where("id = ?", projectID).First(&project).Error; err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
-			s.invalidateCachedComposeProjectID(normalizedName)
+			s.invalidateCachedComposeProjectIDInternal(normalizedName)
 			return nil, false, nil
 		}
 		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-			return nil, false, fmt.Errorf("request canceled or timed out")
+			return nil, false, fmt.Errorf("request canceled or timed out: %w", err)
 		}
 		return nil, false, fmt.Errorf("failed to get project by cached compose name: %w", err)
+	}
+	if normalizeComposeProjectName(project.Name) != normalizedName {
+		s.invalidateCachedComposeProjectIDInternal(normalizedName)
+		return nil, false, nil
 	}
 
 	return &project, true, nil
 }
 
-func (s *ProjectService) rebuildComposeNameCache(ctx context.Context) error {
+func (s *ProjectService) rebuildComposeNameCacheInternal(ctx context.Context) error {
 	var projects []models.Project
 	if err := s.db.WithContext(ctx).Select("id", "name").Find(&projects).Error; err != nil {
 		return err
@@ -318,24 +322,24 @@ func (s *ProjectService) GetProjectByComposeName(ctx context.Context, name strin
 	var proj models.Project
 	err := s.db.WithContext(ctx).Where("name = ? OR name = ?", name, normalized).First(&proj).Error
 	if err == nil {
-		s.cacheComposeProjectID(normalized, proj.ID)
+		s.cacheComposeProjectIDInternal(normalized, proj.ID)
 		return &proj, nil
 	}
 	if !errors.Is(err, gorm.ErrRecordNotFound) {
 		return nil, fmt.Errorf("failed to get project by name: %w", err)
 	}
 
-	if cachedProject, found, cacheErr := s.lookupProjectByCachedComposeName(ctx, normalized); cacheErr != nil {
+	if cachedProject, found, cacheErr := s.lookupProjectByCachedComposeNameInternal(ctx, normalized); cacheErr != nil {
 		return nil, cacheErr
 	} else if found {
 		return cachedProject, nil
 	}
 
-	if err := s.rebuildComposeNameCache(ctx); err != nil {
+	if err := s.rebuildComposeNameCacheInternal(ctx); err != nil {
 		return nil, fmt.Errorf("failed to list projects by compose name: %w", err)
 	}
 
-	if cachedProject, found, cacheErr := s.lookupProjectByCachedComposeName(ctx, normalized); cacheErr != nil {
+	if cachedProject, found, cacheErr := s.lookupProjectByCachedComposeNameInternal(ctx, normalized); cacheErr != nil {
 		return nil, cacheErr
 	} else if found {
 		return cachedProject, nil
--- a/backend/internal/services/project_service_test.go
+++ b/backend/internal/services/project_service_test.go
@@ -325,7 +325,7 @@ func TestProjectService_GetProjectByComposeName(t *testing.T) {
 		assert.Equal(t, display.ID, found.ID)
 	})
 
-	t.Run("invalidates stale normalized cache entries", func(t *testing.T) {
+	t.Run("invalidates stale normalized cache entries after deletion", func(t *testing.T) {
 		db := setupProjectTestDB(t)
 		svc := NewProjectService(db, nil, nil, nil, nil, nil)
 
@@ -340,7 +340,7 @@ func TestProjectService_GetProjectByComposeName(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, original.ID, found.ID)
 
-		cachedProjectID, cached := svc.getCachedComposeProjectID("myproject")
+		cachedProjectID, cached := svc.getCachedComposeProjectIDInternal("myproject")
 		require.True(t, cached)
 		assert.Equal(t, original.ID, cachedProjectID)
 
@@ -357,9 +357,41 @@ func TestProjectService_GetProjectByComposeName(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, replacement.ID, found.ID)
 
-		cachedProjectID, cached = svc.getCachedComposeProjectID("myproject")
+		cachedProjectID, cached = svc.getCachedComposeProjectIDInternal("myproject")
 		require.True(t, cached)
 		assert.Equal(t, replacement.ID, cachedProjectID)
+	})
+
+	t.Run("invalidates stale normalized cache entries after rename", func(t *testing.T) {
+		db := setupProjectTestDB(t)
+		svc := NewProjectService(db, nil, nil, nil, nil, nil)
+
+		original := &models.Project{
+			BaseModel: models.BaseModel{ID: "p5"},
+			Name:      "My App!",
+			Path:      "/tmp/my-app",
+		}
+		require.NoError(t, db.Create(original).Error)
+
+		found, err := svc.GetProjectByComposeName(ctx, "myapp")
+		require.NoError(t, err)
+		assert.Equal(t, original.ID, found.ID)
+
+		cachedProjectID, cached := svc.getCachedComposeProjectIDInternal("myapp")
+		require.True(t, cached)
+		assert.Equal(t, original.ID, cachedProjectID)
+
+		require.NoError(t, db.Model(&models.Project{}).Where("id = ?", original.ID).Updates(map[string]any{
+			"name": "New Service",
+			"path": "/tmp/new-service",
+		}).Error)
+
+		_, err = svc.GetProjectByComposeName(ctx, "myapp")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "project not found")
+
+		_, cached = svc.getCachedComposeProjectIDInternal("myapp")
+		assert.False(t, cached)
 	})
 }
 
--- a/backend/internal/services/updater_service.go
+++ b/backend/internal/services/updater_service.go
@@ -1380,7 +1380,7 @@ func (s *UpdaterService) restartContainersUsingOldIDs(ctx context.Context, oldID
 		}
 	}
 
-	updatedProjectServiceCounts := make(map[string]int)
+	processedProjectServices := make(map[string]map[string]struct{})
 	projectResults := make(map[string]error) // stores error from the project-level update run
 
 	var results []updater.ResourceResult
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by spupuz** [comment]
image

@kmendell these are failing cause are going in timeout... i don't know ho to solve them

**Comment by spupuz** [comment]
@kmendell let me know if you can do something

File: frontend/src/routes/(app)/environments/[id]/components/JobsTab.svelte [hunk 1]
```diff
@@ -268,6 +268,17 @@
...
 												{#if job.id === 'auto-update' && $formInputs.autoUpdate.value}
 													<div class="border-border/20 space-y-3 border-t pt-3">
+														<div class="flex flex-row items-center justify-between rounded-lg border p-4">
+															<div class="space-y-0.5">
+																<Label class="text-base">Use compose update</Label>
+																<p class="text-muted-foreground text-sm">
+																	Apply updates using <code>docker compose pull</code>
+																	and <code>docker compose up -d</code> for containers belonging to a compose project.
+																</p>
+															</div>
```
Comment (kmendell) [inline frontend/src/routes/(app)/environments/[id]/components/JobsTab.svelte]: Missing i18n strings

backend/internal/services/project_service.go
```diff
@@ -199,6 +199,67 @@ func (s *ProjectService) GetProjectFromDatabaseByID(ctx context.Context, id stri
 	return &project, nil
 }
 
+func (s *ProjectService) GetProjectByComposeName(ctx context.Context, name string) (*models.Project, error) {
```
kmendell, line 316: Are all of these functions needed, The logic for getting projects already exists, these seem like duplicate code just to retive projects by name instead.

spupuz:
@kmendell let me know if you can do something


> Few things i found, but im still confused on how this is different fomr the updater depends on. When the updater (currently) runs on a conatnier its copys its exact config, and then creates a new container with that exact config labels, networks, everything, so im not sure what this solves exactly.

Hi kmendell, that's a fair question. You're right that Arcane already attempts to clone the container config, but the new Compose-aware methodology is fundamentally different in terms of how and why it operates.

The key differences are:

Native Orchestration vs Manual Reconstruction: The current standalone methodology manually reconstructs a docker run command by inspecting the container. This is a "best-effort" reconstruction that can be fragile with complex configurations. The new methodology uses the native Docker Compose engine as the source of truth. By calling docker compose pull/up, we offload the orchestration complexity to Docker itself, which is significantly more robust.

Selective & Intelligent Service Updates: While the standalone flow has a custom dependency sorter (in sorter.go), it's still a manual simulation. Using native Compose allows us to run docker compose up -d [services], letting Docker intelligently decide which containers actually need a restart based on image or config divergence. This prevents "blind" restarts of the entire chain if nothing has changed.

Optimized Downtime: Standalone updates always follow a stop -> rm -> run cycle, causing guaranteed downtime. The Compose-aware flow uses RecreateDiverged logic, meaning Docker only recreates what is strictly necessary. For the background Auto-Updater, we've also added explicit SHA256 digest comparisons before pulling to minimize any unnecessary activity.

Resource Integrity: Compose-managed networks, aliases, and volumes are managed more reliably by the native engine than by Arcane's manual reconstruction logic. This ensures that a project container always comes back up exactly as the user (or the project) defined it.

In short: we are moving from Arcane simulating an update to Arcane delegating it to the official tool that manages the project.

frontend/src/routes/(app)/environments/[id]/components/JobsTab.svelte
Comment (spupuz) [reply frontend/src/routes/(app)/environments/[id]/components/JobsTab.svelte]: Commit ID: 75fca387
Branch: feat/compose-update-logic
Updated Files: 

en.json
 and 

JobsTab.svelte

backend/internal/services/project_service.go
spupuz: ### Redundancy Audit: ProjectService

I analyzed the concern about potential code duplication in [ProjectService](cci:2://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/services/project_service.go:34:0-40:1). While we do have existing methods to list and retrieve projects, [GetProjectByComposeName](cci:1://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/services/project_service.go:201:0-215:1) is specifically required for the **Compose-aware update logic** for the following reasons:

1.  **Bridging Docker State to Database**: 
    When the [UpdaterService](cci:2://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/services/updater_service.go:27:0-42:1) inspects a container, it only sees Docker labels like `com.docker.compose.project`, which provide the project name as a **string**. Since Arcane uses internal **UUIDs**, we need a direct bridge to find the corresponding database record using that string to retrieve the project's filesystem path and local ID.

2.  **Handling Compose Normalization**: 
    Docker Compose applies specific normalization rules (e.g., lowercase, removing special characters) to project names via `loader.NormalizeProjectName`. This function handles that logic explicitly, checking both the raw label name and the normalized version to ensure we find the correct project regardless of how it was originally named or stored.

3.  **Service-Layer Efficiency**: 
    Functions like [ListProjects](cci:1://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/huma/handlers/projects.go:323:0-371:1) are optimized for UI pagination; they use `LIKE` queries, return complex DTOs, and trigger concurrent live status fetching for every item. [GetProjectByComposeName](cci:1://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/services/project_service.go:201:0-215:1) is a "lean" service-layer method designed for background tasks, returning the raw model directly and avoiding the overhead of UI-centric logic.

Consistent with other services (like [EnvironmentService](cci:2://file:///c:/Users/Belloni%20Alessandro/NextCloud/Various/DockerSource/arcane/backend/internal/services/environment_service.go:28:0-34:1)), having explicit `GetBy[Property]` methods ensures we have reliable, typed retrieval for core logic without relying on broad search functions.

spupuz:
@kmendell i'm i supposed to do something or are you fixing the conflicts?

kmendell:
yes youll have to fix the conflicts im too busy at the moment to do that and finish up the other stuff.

spupuz:
@kmendell i should have fixed the merge problems but still the E2E test are going long.

spupuz:
@kmendell alla seems fine now

kmendell:
I fixed the test the other day. Ill try to look at it again soon. Trying to figure out these bugs first.

kmendell:
@user00265 Do you want to look at this as a first glance?

spupuz:
@kmendell i'm keeping rebasing and merging conflict ... let me know when you are ready to merge this so i'll do the work once for all.

kmendell:
@spupuz you can fix them now and ill look again.

spupuz:
@kmendell hope this works now

kmendell:
IF this is going to be implmented, i think getting of of the selector in the UI and just making the backend handle it is the direction to do , if it has teh compose labels then use them. There shouldnt need to be a UI element in that regard i dont think.

spupuz:
> IF this is going to be implmented, i think getting of of the selector in the UI and just making the backend handle it is the direction to do , if it has teh compose labels then use them. There shouldnt need to be a UI element in that regard i dont think.


@kmendell I completely agree with removing the UI selector and making the backend handle it automatically when Compose labels are present. I'll update the PR to reflect this.

To address the update logic itself and why it is implemented this way: while not everyone uses Compose labels (so standard containers will respectfully fall back to the normal inline update), the Compose update logic is extremely valid and necessary to avoid problems for containers that do belong to a compose project. It allows for a much more correct management of Compose stacks.

If we look at the standard inline update path in 

updater_service.go (updateContainer), it performs a raw ContainerRemove and ContainerCreate using the Docker API with a copied HostConfig. This causes several limitations for Compose-managed containers:

It ignores depends_on relationships, healthchecks, and lifecycle hooks defined in docker-compose.yml.
It struggles with networking: Arcane has to manually sanitize configurations (like removing memorySwappiness or stripping endpoint mac addresses) to avoid creation errors, which can cause the container to drift from its declarative configuration.
Updating a single container inline leaves the docker-compose internal state detached.
By hooking into ProjectService to trigger the actual docker/compose/v5/pkg/api commands (essentially docker compose pull and docker compose up -d), we guarantee that the compose project correctly pulls the new image and re-evaluates the entire stack cleanly, respecting dependencies and networks as defined in the source of truth.

I'll proceed to drop the UI toggle and automate the backend trigger based on the labels!

kmendell:
The the container are deployed from a docker compose project docker compsoe automatcially adds com.docker.compose labels to it, arcane replciates this functionality as well.

spupuz:
> The the container are deployed from a docker compose project docker compsoe automatcially adds com.docker.compose labels to it, arcane replciates this functionality as well.

so how we do proceed it's been 2 week that i'm keeping rebasing and recommiting 👯‍♂️

PR #2198, opened 2026-03-31
fix: resolve project status using effective compose project name

### LINKED ISSUE #1746: 🐞 Bug: Acrane states that project is stopped but it's up
### Bug Description

Acrane states that project is stopped but it's up.
the project is https://github.com/mailcow/mailcow-dockerized
docker compose file is unchanged

### Steps To Reproduce

Bug description and screenshots are self-explanatory

### Expected Behavior

Should show a project as running and healthy

### Actual Behavior

Bug description and screenshots are self-explanatory

### lazydocker
Image

### Projects page
Image

### Containers page
Image

### Arcane Version

1.15.0

### Installation Method

Docker Compose (Recommended)

### Environment Type

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

28.5.1-ce

### Browser & Version

firefox

```shell

```

```yaml

```

```shell

```

### Additional Context

_No response_

GiulioSavini:
## Summary
- Projects with `COMPOSE_PROJECT_NAME` in their `.env` were shown as "stopped" on the Projects page while actually running
- Root cause: container-to-project matching used the normalized directory name, but Docker labels containers with the effective compose project name from `COMPOSE_PROJECT_NAME`
- Fix: store the effective compose project name in the DB and use it as fallback for container matching

Fixes #1746

## Root cause

When Docker Compose starts containers, it labels them with `com.docker.compose.project` = the effective project name. This name comes from (in priority order):
1. `COMPOSE_PROJECT_NAME` environment variable / `.env` file
2. The directory name

Arcane matched containers to projects using only `normalizeComposeProjectName(directoryName)`. For mailcow-dockerized, the directory is `mailcow-dockerized` (normalized: `mailcow-dockerized`), but the `.env` sets `COMPOSE_PROJECT_NAME=mailcowdockerized` (no hyphen). The names don't match → zero containers found → project shown as stopped.

The Containers page was immune because it lists containers directly from Docker without project-name matching.

## Changes

**`backend/internal/models/project.go`**
- Added `ComposeProjectName *string` field to `Project` model

**`backend/resources/migrations/{sqlite,postgres}/045_add_compose_project_name.{up,down}.sql`**
- Migration to add/remove the `compose_project_name` column

**`backend/internal/services/project_service.go`**
- Added `resolveComposeProjectNameInternal`: loads the compose file without forcing a project name, letting compose-go resolve it from `COMPOSE_PROJECT_NAME` or directory name. Returns the name only if it differs from the normalized directory name.
- Added `ptrStringEqualInternal`: helper for nullable string comparison
- `upsertProjectForDir`: populates `ComposeProjectName` during project discovery and sync
- `mapProjectToDto`: falls back to `ComposeProjectName` when normalized name finds no containers
- `GetProjectStatusCounts`: same fallback

## Test plan
- [x] `go test ./internal/services/...` passes
- [ ] Create a project with `COMPOSE_PROJECT_NAME=customname` in `.env` and a directory name that differs
- [ ] Start the project via Docker Compose CLI
- [ ] Verify the Projects page shows "Running" (not "Stopped")
- [ ] Verify projects without `COMPOSE_PROJECT_NAME` still match correctly
- [ ] Verify the Containers page still works as before

This PR fixes a long-standing bug where projects using `COMPOSE_PROJECT_NAME` in their `.env` file were always displayed as "Stopped" even when their containers were running. The root cause was that container-to-project matching relied solely on the normalised directory name, while Docker labels containers with the effective compose project name (which `COMPOSE_PROJECT_NAME` can override).

**Changes:**
- Adds a `ComposeProjectName *string` field to the `Project` model and a corresponding migration (045) for both SQLite and Postgres.
- Introduces `resolveComposeProjectNameInternal`: loads the compose project with an empty name so compose-go resolves it from the `.env` file; stores the result only when it differs from the normalised directory name.
- Updates `upsertProjectForDir` to populate/sync `ComposeProjectName` on every discovery/sync pass.
- Updates both `mapProjectToDto` and `GetProjectStatusCounts` with an identical two-step lookup: try the normalised directory name first, fall back to `ComposeProjectName` only when no containers are found.
- Adds `ptrStringEqualInternal` for nullable-string comparison used in the sync diff check.

The overall approach is sound and correctly follows the compose name-resolution priority order. The SQLite down migration properly reconstructs the full table (required because SQLite doesn't support `DROP COLUMN` in older versions) and its column list matches the schema established by migration 044.
</details>

- Safe to merge; the fix is correct and fails gracefully when the compose file cannot be loaded.
- All remaining findings are P2 style/performance suggestions. The logic is correct: the two-step lookup is applied consistently in both `mapProjectToDto` and `GetProjectStatusCounts`, the migration is sound for both SQLite and Postgres, and the fallback is gated behind a nil-check so existing projects without `COMPOSE_PROJECT_NAME` are unaffected.
- backend/internal/services/project_service.go — double compose-file load per project per sync is worth addressing before it becomes a bottleneck on large installations.
</details>

| Filename | Overview |
|----------|----------|
| backend/internal/services/project_service.go | Adds `resolveComposeProjectNameInternal` and `ptrStringEqualInternal` helpers; updates `upsertProjectForDir`, `mapProjectToDto`, and `GetProjectStatusCounts` to store and use the effective compose project name as a fallback for container-matching. Logic is correct but causes a double compose-file load per project on every sync. |
| backend/internal/models/project.go | Adds optional `ComposeProjectName *string` field with correct GORM column tag and JSON omitempty; straightforward model change. |
| backend/resources/migrations/postgres/045_add_compose_project_name.up.sql | Adds nullable TEXT column `compose_project_name` to the `projects` table; minimal and correct. |
| backend/resources/migrations/sqlite/045_add_compose_project_name.down.sql | Full table-recreate rollback (required for SQLite column drops); column list matches migration 044's schema, indexes recreated correctly. |

</details>

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[upsertProjectForDir] --> B[countServicesFromCompose\nload with normalized dir name]
    A --> C[resolveComposeProjectNameInternal\nload with empty project name]
    C --> D{effectiveName equals\nnormalized dirName?}
    D -- Yes --> E[return nil - no override needed]
    D -- No --> F[return pointer to effectiveName]
    F --> G[ComposeProjectName persisted to DB]

    H[mapProjectToDto or GetProjectStatusCounts] --> I[Primary lookup\ncontainersByProject normalized-name]
    I --> J{containers found?}
    J -- Yes --> K[Use matched containers]
    J -- No --> L{ComposeProjectName set\nand differs from normName?}
    L -- Yes --> M[Fallback lookup\ncontainersByProject ComposeProjectName]
    L -- No --> N[No containers - Stopped or Unknown]
    M --> K
```
</details>

`````markdown
Path: backend/internal/services/project_service.go
Line: 3226

Comment:
**Pointer idiom inconsistency**

The rest of this file creates `*string` values using `new(value)` (e.g., `new(dirName)`, `new(reason)`, `new("healthy")`), but this new function uses `return &effectiveName`. Per the project's pointer-initialisation convention, prefer the consistent form:

```suggestion
	return new(effectiveName)
```

**Rule Used:** What: Use the new built-in `new` function to initi... ([source](https://app.greptile.com/review/custom-context?memory=4da04955-3abb-4e9c-ae2f-26c1a1adeb57))

---

Path: backend/internal/services/project_service.go
Line: 893-894

Comment:
**Compose file loaded twice per project per sync**

`countServicesFromCompose` (line 893) and `resolveComposeProjectNameInternal` (line 894) both independently call `LoadComposeProjectFromDir`, each also re-fetching the projects directory setting and path mapper from the settings service. For installations with many projects this doubles the compose-file parsing work on every sync cycle.

Consider extracting a shared `loadedProject` struct (or accepting the loaded `*composetypes.Project` as a parameter) so the file is parsed once and both pieces of information are derived from the same parse result.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: resolve project status using effect..."](https://github.com/getarcaneapp/arcane/commit/1a60b82e5c6ac5eae973fdaaa6b2f3caf3c10e92) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26881026)</sub>

> Greptile also left **2 inline comments** on this PR.

### CODE REVIEW

**Comment by kmendell** [comment]
I dont think this needs a whole new column in the table just to a "effective name" this is more over some validation logic failing somewhere or doing the wrong thing. Ill try to look tonight.

**Comment by GiulioSavini** [comment]
The issue isn't really validation logic — the matching works correctly for what it knows. The problem is that normalizeComposeProjectName(directoryName) is the only key used to look up containers, but Docker labels them with
  whatever COMPOSE_PROJECT_NAME resolves to. When .env sets COMPOSE_PROJECT_NAME=mailcowdockerized and the directory is mailcow-dockerized, there's no way to match them without reading the compose config to get the effective    
  name.

**Comment by kmendell** [comment]
right, so what im saying is if docker for some reason dowesnt allow dashes (which the do im pretty sure), then why is the normalize function removing it, this also bring up the other bug ive forgotten to fix is i dont think the "name" filed in a compose yaml is respected as the project name when it should be.

**Comment by kmendell** [comment]
The project name is set here technically speaking:
https://github.com/getarcaneapp/arcane/blob/4d1fabbda8579173532f914399d38cd996845c80/backend/pkg/projects/load.go#L140

**Comment by kmendell** [comment]
Or rather that where the label comes from or should if its deployed with arcane.

**Comment by GiulioSavini** [comment]
well...You're right about the name field — that's a real bug. Arcane always forces normalizeComposeProjectName(directoryName) as the project name when calling LoadComposeProjectFromDir (load.go L140), which overrides whatever compose-go would resolve from the name: field in the YAML. However, the normalize function itself doesn't strip dashes — loader.NormalizeProjectName("mailcow-dockerized") returns mailcow-dockerized unchanged. The mismatch in #1746 comes from the fact that mailcow's own .env sets COMPOSE_PROJECT_NAME=mailcowdockerized (no hyphen), so Docker labels the containers with mailcowdockerized  while Arcane looks them up as mailcow-dockerized. Both cases (COMPOSE_PROJECT_NAME override and name: field) share the same root cause: Arcane forces the directory name instead of letting compose-go resolve theeffective project name. My resolveComposeProjectNameInternal handles both — it passes an empty project name so compose-go resolves from COMPOSE_PROJECT_NAME, name: field, or directory name in that order, matching what Docker actually does. That said, if you'd prefer to fix this at the LoadComposeProjectFromDir level (not forcing the name) rather than adding the fallback column, I'm happy to rework the approach. Just let me know which direction you'd prefer. Appreciate your work :)

**Comment by kmendell** [comment]
yes i dont feel we need a brand new column for this , id rather fix whats broken, the name field from the ui, mainly should just serve as the folder name, or project name if none if provided id think is what iwas going for.

**Comment by GiulioSavini** [comment]
Right, I'll close this pr. Do you have something you would like or need me to work on? Bug features etc. Something that you prioritize. Thnks

**Comment by kmendell** [comment]
You can work on anything you would like, but no pressure on anything, if you want to try and do this one still you can, dont matter to me :)

**Comment by kmendell** [comment]
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/kmendell-XqXhZ889F7MYCfD6CFnDWH/pr-checks/ecb241d4-95fc-42a6-a609-9e3f8732a14b?refs=e54043a7eea2e1fde08ead9d9a64282cdd07da1e&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

**Comment by GiulioSavini** [comment]
thanks for the sync

File: backend/internal/services/project_service.go [hunk 1]
```diff
@@ -242,6 +242,16 @@ func buildProjectImagePullPlan(services composetypes.Services) map[string]imageP
 	return plan
 }
 
+func ptrStringEqualInternal(a, b *string) bool {
```
kmendell: What is this function used for this seems to be dead code?

GiulioSavini: It's used just a few lines below — in `upsertProjectForDir` — to guard the `compose_project_name` update:

```go
if !ptrStringEqualInternal(existing.ComposeProjectName, composeProjectName) {
    updates["compose_project_name"] = composeProjectName
}
```

It handles the nil-nil / nil-nonnil / value comparison for `*string` without duplicating that logic inline. Happy to inline it if you'd prefer to avoid a named helper for a single callsite.

kmendell (approved):

Issue #1052, opened 2025-12-03
🐞 Bug: System Overview has wrong reference for CPU Cores and RAM


Smoerrebroed:
### Bug Description

In the main dashboard, I see the following:

<img width="1027" height="304" alt="Image" src="https://github.com/user-attachments/assets/51fc2436-7f18-473f-9748-718b5b446a50" />

Regardless of how much the CPUs are used, the point of reference here is 2 cores ("2 Prozessoren") because that is what I allocated to the container running Arcane. Likewise, I allocated 512 MB of RAM. However, I believe that in the general overview, the total CPU utilization as well as the total RAM usage (for the whole machine) is displayed. (As you can see: I'm using 4.69 GB out of the 512 MB allocated.) With that in mind, the reference/upper ceiling should be based on the amount of physical RAM and CPU cores for the host.

### Steps To Reproduce

Install Arcane 1.10.0 in a container
Log in and look at the main dashboard

### Expected Behavior

The point of reference both for CPU cores and RAM on the main dashboard should be what's available to Docker, not only what's available to the container running Arcane.

### Actual Behavior

The point of reference (physical) is taken from the resource allocation for the Arcane container - which doesn't make much sense IMHO.

### Screenshots

<img width="1027" height="304" alt="Image" src="https://github.com/user-attachments/assets/ec4b8e6c-3459-449e-86bb-cb601197a2d7" />

### Arcane Version

1.10.0

### Operating System

Linux

### Browser

Brave 1.84.141

### Docker Version

25.3.0

kmendell:
This will be fixed in 1.10.1 https://github.com/getarcaneapp/arcane/commit/80ccc9b689afa7fa1e2ac87862970652772153cd

Smoerrebroed:
Can't confirm. I just now updated to 1.10.1 and it sill shows the values related to the Docker container that Arcane is running in. What now?

Issue #995, opened 2025-11-27
🐞 Bug: Unexpected containers update behavior.


ThreeTheOG:
### Bug Description

When I update for example my media stack, or immich It restarts nonfunctional. I then go into the ui and restart it, same issue. I then ssh into the server and manually docker compose down && docker compose up -d and it everything is fine and dandy. I don't know if there is somehow a permissions issue, or there is something with the launch commands. Would prefer not to share all of my compose yamls.

### Steps To Reproduce

Go to projects
click update projects

### Expected Behavior

Project goes online

### Actual Behavior

Is borked.

### Arcane Version

1.9.0

### Operating System

Linux

### Browser

zen(firefox) latest

### Docker Version

27.1.1

kmendell:
The updater restarts the container not the entire project defintion, but this should matter it retains its exisiting config. Can you provide a bit more insight on what you mean by non functional?

ThreeTheOG:
I believe Gluetun restarts unhealthy, and others are set to stop or not start if until gluetun is healthy. I have had issues with other projects as well, I'm assuming its something networking related. All projects that I'm having issues with are multi-container i believe. Will mess with it some more when I have time or if you have a hunch.

kmendell:
@ThreeTheOG Can you try this image? `ghcr.io/getarcaneapp/arcane:compose-restart`

ThreeTheOG:
I just ran that version and ran into an odd issue. When I clicked “Update Projects” from the Projects page, the stack failed to start. Gluetun showed as healthy, but about half of the containers were marked as not created.

I SSH’d into the VM and ran my alias command:

`docker compose pull && docker compose down && docker compose up -d`

It ended up pulling four images in the stack, and after that everything came up fine. So I’m assuming the issue was that Arcane didn’t pull the updated images, but I’m not sure. Is there any kind of log I can check for image-pull errors from Arcane?

kmendell:
@ThreeTheOG If you set LOG_LEVEL=debug in the env and the redeploy arcane that should show you alot more information.

kmendell:
@ThreeTheOG Can you try this image please? `ghcr.io/getarcaneapp/arcane:updater-restart`

stavros-k:
[...]

PR #2343, opened 2026-04-11
fix: show host CPU/RAM in System Overview instead of Arcane container limits (#1110)

### LINKED ISSUE #1110: 🐞 Bug: System Overview has wrong reference for CPU Cores and RAM (#1052)
### Bug Description

It seems that the solution mentioned in https://github.com/getarcaneapp/arcane/issues/1052 does not fix the issue.

### Steps To Reproduce

See https://github.com/getarcaneapp/arcane/issues/1052

### Expected Behavior

See https://github.com/getarcaneapp/arcane/issues/1052

### Actual Behavior

See https://github.com/getarcaneapp/arcane/issues/1052

### Screenshots

Image

Image

### Arcane Version

1.10.1

### Operating System

Linux

### Browser

Brave

### Docker Version

28.3.1

### Additional Context

_No response_

GiulioSavini:
## Summary

Fixes #1110 (originally reported as #1052). The System Overview dashboard reports CPU cores and memory totals clamped to the Arcane container's own cgroup limits. When Arcane runs in a container with e.g. `cpus: 2` and `memory: 512MB`, the dashboard shows things like `Memory Usage 4.69 GB / 512 MB` — `memUsed` is read from the host while `memTotal` has been overwritten with the container limit — and `2 CPUs` even when the host has many more. The referenced earlier fix (80ccc9b...) did not remove the override that caused the regression on both code paths that feed the dashboard.

## Fix

Two surgical deletions, one per path, both leaving the surrounding infrastructure intact.

1. **`backend/internal/api/ws_handler.go`** — drop the `applyCgroupLimits` call from `collectSystemStats` and delete the now-orphaned helper. `collectSystemStats` now returns what gopsutil reports for the host (`cpu.Counts(true)`, `mem.VirtualMemory()`) unchanged.
   - The cgroup-limits caching infrastructure (`cgroupLimitsDetector`, `getCachedCgroupLimitsInternal`, the static cache sampler) stays put — still exercised by `TestWebSocketHandler_GetCachedCgroupLimitsInternal_DeduplicatesRefresh`. If someone wants to add a future "Arcane container stats" view that *intentionally* shows the container's own limits, the infra is there to consume; the regression just isn't wired into the host dashboard any more.
2. **`backend/internal/huma/handlers/system.go`** — drop the `DetectCgroupLimits` block in `GetDockerInfo` and its now-unused `dockerutil` import. `info.NCPU` and `info.MemTotal` from the Docker daemon (which already reflect the host the daemon is running on) are passed through unchanged.

A short comment explains the intent in `GetDockerInfo` so a future refactor doesn't re-introduce the clamp.

## Why this is safe

- `applyCgroupLimits` had **one** caller (`collectSystemStats`) — confirmed by repo-wide grep. The helper is deleted cleanly, no dangling references.
- The `docker "github.com/getarcaneapp/arcane/backend/pkg/dockerutil"` alias in `system.go` was **only** used inside the removed cgroup block — confirmed by the same grep. Removing it keeps imports tidy.
- The `dockerutil.DetectCgroupLimits` function itself and the whole `backend/pkg/dockerutil` package are untouched.
- No frontend changes. The dashboard will simply start rendering the host values it was always fetching on that path.

## Tests

- `go vet ./internal/api/... ./internal/huma/handlers/...` — clean.
- `go test ./internal/api/... -count=1` — `ok` in ~3.4s. All existing `ws_handler` tests pass, including the one that exercises `getCachedCgroupLimitsInternal` (which is still around and still reachable).
- `go test ./internal/huma/handlers/... -count=1` — `ok` in ~0.2s.

## Out of scope

- `#995` (container update behavior) — intentionally not touched here, that one is a design discussion and `kmendell`'s `compose-restart` / `updater-restart` experimental images already explore it.
- Any future "show me my Arcane container's own resource budget" view — this would be a separate, additive PR that consumes the still-intact cgroup infrastructure.

Fixes #1110

This PR fixes a long-standing dashboard regression (#1110) where System Overview CPU/RAM totals were clamped to Arcane's container cgroup limits instead of showing real host values. The approach is more nuanced than the PR description suggests: rather than deleting the cgroup-limits logic, it retains it for LXC deployments (where cgroup limits represent the true guest budget) while short-circuiting it when running inside Docker (where `--cpus`/`--memory` are artificial operator caps). The PR also sneaks in two unrelated but clearly correct fixes: a race condition in `ws_proxy.go` (register the pending stream before sending the start message) and a log-stream ordering fix in `ws_handler.go` (mark done before broadcasting the error, preventing a race where a new subscriber could join a dying stream).
</details>

Safe to merge — the core fix is correct, the LXC fallback is sound, and the ancillary changes improve correctness.

All findings are P2. The only actionable note is caching IsDockerContainer() to avoid repeated syscalls on every stats tick, which is a minor performance improvement rather than a correctness issue.

backend/pkg/dockerutil/cgroup_utils.go — IsDockerContainer should cache its result.
</details>

<a href="https://chatgpt.com/codex/deeplink?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22getarcaneapp%2Farcane%22%20on%20the%20existing%20branch%20%22fix%2Fsystem-overview-use-host-stats%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22fix%2Fsystem-overview-use-host-stats%22.%0A%0AFix%20the%20following%201%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Abackend%2Fpkg%2Fdockerutil%2Fcgroup_utils.go%3A287-296%0A**Cache%20the%20result%20%E2%80%94%20result%20is%20constant%20for%20process%20lifetime**%0A%0A%60IsDockerContainer%28%29%60%20performs%20an%20%60os.Stat%28%22%2F.dockerenv%22%29%60%20and%20potentially%20an%20%60os.ReadFile%28%22%2Fproc%2Fself%2Fcgroup%22%29%60%20on%20every%20call.%20In%20the%20WebSocket%20stats%20path%2C%20%60applyCgroupLimits%60%20calls%20this%20function%20on%20every%20stats%20tick%20%28e.g.%2C%20once%20per%20second%29%2C%20so%20these%20syscalls%20repeat%20indefinitely%20even%20though%20the%20answer%20can%20never%20change%20during%20the%20process%20lifetime.%0A%0AConsider%20a%20package-level%20%60sync.Once%60%20to%20evaluate%20and%20cache%20the%20result%20once%3A%0A%0A%60%60%60go%0Avar%20%28%0A%20%20%20%20isDockerContainerOnce%20%20%20sync.Once%0A%20%20%20%20isDockerContainerResult%20bool%0A%29%0A%0Afunc%20IsDockerContainer%28%29%20bool%20%7B%0A%20%20%20%20isDockerContainerOnce.Do%28func%28%29%20%7B%0A%20%20%20%20%20%20%20%20if%20_%2C%20err%20%3A%3D%20os.Stat%28%22%2F.dockerenv%22%29%3B%20err%20%3D%3D%20nil%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20isDockerContainerResult%20%3D%20true%0A%20%20%20%20%20%20%20%20%20%20%20%20return%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20data%2C%20err%20%3A%3D%20os.ReadFile%28%22%2Fproc%2Fself%2Fcgroup%22%29%0A%20%20%20%20%20%20%20%20if%20err%20!%3D%20nil%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20return%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20isDockerContainerResult%20%3D%20cgroupV1ContainerPattern.Match%28data%29%20%7C%7C%20cgroupV2ContainerPattern.Match%28data%29%0A%20%20%20%20%7D%29%0A%20%20%20%20return%20isDockerContainerResult%0A%7D%0A%60%60%60%0A%0A"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=2">Fix All in Codex</picture></a>

`````markdown
Path: backend/pkg/dockerutil/cgroup_utils.go
Line: 287-296

Comment:
**Cache the result — result is constant for process lifetime**

`IsDockerContainer()` performs an `os.Stat("/.dockerenv")` and potentially an `os.ReadFile("/proc/self/cgroup")` on every call. In the WebSocket stats path, `applyCgroupLimits` calls this function on every stats tick (e.g., once per second), so these syscalls repeat indefinitely even though the answer can never change during the process lifetime.

Consider a package-level `sync.Once` to evaluate and cache the result once:

```go
var (
    isDockerContainerOnce   sync.Once
    isDockerContainerResult bool
)

func IsDockerContainer() bool {
    isDockerContainerOnce.Do(func() {
        if _, err := os.Stat("/.dockerenv"); err == nil {
            isDockerContainerResult = true
            return
        }
        data, err := os.ReadFile("/proc/self/cgroup")
        if err != nil {
            return
        }
        isDockerContainerResult = cgroupV1ContainerPattern.Match(data) || cgroupV2ContainerPattern.Match(data)
    })
    return isDockerContainerResult
}
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: mark log stream done before broadca..."](https://github.com/getarcaneapp/arcane/commit/ded7b2be9a7989625d23a839a32df439a0b5e864) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28083940)</sub>


kmendell:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/kmendell-XqXhZ889F7MYCfD6CFnDWH/pr-checks/f8e3804f-3371-4780-981f-022c502b3d8c?refs=b824ba0632a9690322021c6a6bd9c6f251e90295&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

kmendell:
The issue with this is LXC containers need the cgroup logic in order to detect the correct limits.

GiulioSavini:
Good catch, @kmendell. The previous version removed `applyCgroupLimits` entirely, which broke LXC deployments where the cgroup limits represent the real hardware budget assigned to the guest (gopsutil reads the host's `/proc` there and would show the physical machine's full RAM/CPU).

I've reworked the fix to be environment-aware:

- Added `IsDockerContainer()` to `cgroup_utils.go` — checks for `/.dockerenv` (always created by Docker, never by LXC) with a `/proc/self/cgroup` pattern fallback.
- `applyCgroupLimits()` returns early (no-op) when inside Docker, so gopsutil's already-correct host values are passed through unchanged.
- For LXC and other cgroup-managed environments the limits are still applied as before.
- Same guard in `GetDockerInfo` — Docker daemon's `NCPU`/`MemTotal` are already host values in Docker; in LXC the daemon may report the physical machine so the LXC cgroup budget must be applied.

Both `go test ./internal/api/...` and `./internal/huma/handlers/...` pass.

GiulioSavini:
Two things I'd like your eyes on before this merges:

**1. `GetDockerInfo` + LXC: cgroup limits might be double-wrong**

If Arcane and Docker Engine are both running inside the *same* LXC guest (the common case), `info.NCPU`/`info.MemTotal` from the daemon already reflect the LXC-limited view — applying cgroup limits on top could produce incorrect results a second time. The guard `!IsDockerContainer()` correctly skips this in Docker, but for LXC-hosted-Docker the block may be redundant or harmful.

The original PR removed this block entirely for `GetDockerInfo` and I think that's actually correct: the Docker daemon always reports what its own environment sees, which is already the right thing to display. I can drop the cgroup block from `GetDockerInfo` unconditionally and only keep the `applyCgroupLimits` guard (with the `!IsDockerContainer()` check) in `ws_handler.go` where gopsutil reads raw `/proc`. Would you prefer that?

**2. `/.dockerenv` + Podman docker-compat mode**

Podman can create `/.dockerenv` when running in Docker-compatibility mode. In that case `IsDockerContainer()` would return `true` even though Podman's cgroup limits behave more like Docker's (artificial throttles), so the behavior would still be correct — but worth flagging in case you have Podman users who rely on the LXC path.

kmendell:
I will have to test on a LXC container later today

GiulioSavini:
alrght : )

kmendell (approved):

PR #2467, opened 2026-04-29
fix(edge): prevent slog-gin panic on tunneled requests

lohrbini:
The tunnel client passed the full production gin router as its internal http.Handler. slog-gin unconditionally wraps gin's ResponseWriter with a body-capture tee writer on every request. When huma serialises a response via io.Copy, it reads through slog-gin's internal *bytes.Buffer, which is uninitialised against the synthetic responseRecorder used by the tunnel client, causing a nil pointer dereference panic.

## Checklist

- [X ] This PR is **not** opened from my fork’s `main` branch

## What This PR Implements

The fix addresses two bugs that cause failures when applying resources through the edge agent:

* Any request routed through the edge tunnel that passes through huma's response path causes the agent to panic — the agent crashes and recovers, meaning the operation silently fails on the client side.

* Streaming responses over the gRPC transport lose their body — the manager-side collector resolves the response too early and discards all subsequent data chunks, returning a truncated or empty response.

Fixes: 

#### `backend/internal/bootstrap/router_bootstrap.go`
`setupRouter` now builds two routers with identical route registrations. The public-facing `router` retains slog-gin for observability. A new `internalRouter` omits slog-gin and is returned alongside the main router for use by the tunnel client.

**Why:** slog-gin unconditionally wraps gin's `ResponseWriter` with a body-capture tee writer on every request, regardless of filter configuration. The tunnel client serves requests through a synthetic `responseRecorder` rather than a real TCP connection. When huma serialises its response it calls `io.Copy` through the slog-gin-wrapped writer chain, which reads from an uninitialised `*bytes.Buffer` — causing a nil pointer dereference panic on every request that passes through huma's response path.

#### `backend/internal/bootstrap/bootstrap.go`
`setupRouter` now returns `internalRouter` in addition to `router`. `internalRouter` is passed to `StartTunnelClientWithErrors` instead of the main router, ensuring tunneled requests are never handled by a router that includes slog-gin's body-capture writer.

**Why:** The tunnel client previously received the full production router, exposing all tunneled requests to the slog-gin panic described above. Separating the two routers at the call site isolates the fix without affecting the public-facing server.

#### `backend/pkg/libarcane/edge/proxy.go`
`grpcResponseState.handleResponse` now checks `X-Arcane-Tunnel-Stream` before returning early. Previously, if a streaming response envelope arrived with body bytes, the collector resolved immediately and discarded all subsequent `MessageTypeStreamData` chunks. The fix ensures streaming responses always wait for `MessageTypeStreamEnd` before resolving.

**Why:** The agent sends streaming responses in three phases — a `MessageTypeResponse` envelope with headers, one or more `MessageTypeStreamData` chunks with the body, and a final `MessageTypeStreamEnd`. Without the streaming flag check, the collector could resolve on the first envelope and silently drop the body, returning a truncated or empty response to the caller.

## Testing Done

- [X] Development environment started: `./scripts/development/dev.sh start`
- [X] Frontend verified at http://localhost:3000
- [X] Backend verified at http://localhost:3552
- [X] Manual testing completed (describe):
- [X] No linting errors (e.g., `just lint all`)
- [X] Backend tests pass: `just test backend`

AI Tool: Claude
Assistance Level: Significant
What AI helped with: created a patch to fix issues using a terraform provider
I reviewed and edited all AI-generated output: 
I ran all required tests and manually verified changes: tested all changes on multiple edge agents and the manager instance with productive workload

## Additional Context

This PR fixes two bugs in the edge agent: a nil pointer dereference panic caused by slog-gin's body-capture writer wrapping a synthetic `responseRecorder`, and streaming responses being truncated due to early resolution in `collectGRPCResponse`. The slog-gin fix uses a single router with a context-aware middleware that bypasses slog-gin for internal tunnel requests (rather than the two-router approach described in the PR description), which is a cleaner solution. The streaming fix in `handleResponse` correctly defers resolution to `MessageTypeStreamEnd` when `X-Arcane-Tunnel-Stream: 1` is present.
</details>

Safe to merge — both fixes are targeted, well-tested, and don't alter observable behaviour for non-tunnel requests.

No P0 or P1 findings. The implementation correctly addresses both described bugs, the context-key approach is idiomatic Go, the streaming header is always set in canonical form so the direct map lookup is safe, and the new test validates the slog-gin bypass end-to-end.

No files require special attention.
</details>

<sub>Reviews (2): Last reviewed commit: ["remove duplicate router, replace with a ..."](https://github.com/getarcaneapp/arcane/commit/e2f6809e384593c13e18a7027a848148c7a698fd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30109660)</sub>


kmendell:
I reverted the duplciate router, as that causes more performance issues, but i make a filter taht edge requests are skipped by slog-gin. Can you test that out to make sure that fixes it for you?

lohrbini:
I tested your change and build new images. It solves my problem

kmendell (approved):

PR #2999, opened 2026-06-20
feat: allow use of remote trivy server, and only show fixable cves

### LINKED ISSUE #2917: ⚡️ Feature: Option to run trivy in client mode
### Feature Description

Hi @kmendell 

I recently opened an issue https://github.com/getarcaneapp/arcane/issues/2642 but the problem is trivy on arm v7 (32bit). 
I've opened trivy issue and got response:
```
This isn't a physical RAM issue — it's exhaustion of the 32-bit virtual address space when memory-mapping the DB.

Root cause
The error failed to open db: cannot allocate memory comes from bolt.Open. When opening the database, bbolt mmaps the entire trivy.db file into the process's address space.

bbolt's mmapSize rounds the file size up (doubling up to 1GB, then in 1GB steps) and caps it at a per-architecture MaxMapSize:

armv7 (GOARCH=arm): MaxMapSize = 0x7FFFFFFF = 2GB
amd64 / arm64: 0xFFFFFFFFFFFF = 256TB
The uncompressed Trivy DB (the bbolt file) is now larger than ~1GB. On armv7, bbolt tries to mmap a region rounded up to 1–2GB as a single contiguous block. A 32-bit process only has ~2–3GB of virtual address space total, part of which is already taken by the Go runtime heap and other mappings — so there's no contiguous free block of that size, and the kernel returns ENOMEM. On 64-bit (amd64/arm64) the address space is effectively unlimited, so the same file maps fine.

Workaround
The practical workaround is client/server mode: run trivy server on a 64-bit host (the DB lives there and is opened there), and scan from the armv7 board with:

trivy image --server <server-url> adguard/adguardhome:latest
That way the 32-bit client never opens the local bbolt DB.
```

This means that trivy on armv7 is basically useless now. 

Would it be possible to add an option to run trivy in client mode? This would be just switch in the command, for example:
`trivy image --server http://trivy.example.com:10000 adguard/adguardhome:latest`

In a way this makes sense if there are multiple agent, to have a single trivy server and connect all clients to it when scanning. 

### Problem It Solves

Have a centralised trivy server and connect all agents to it.

### Proposed Solution

Add option to run trivy in a client mode and specify server address.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Would you be willing to work on this feature?

- [ ] Yes, I'd like to implement this feature
- [ ] I could help with parts of this feature
- [x] No, I'm just suggesting the feature

kmendell:
## Checklist

## What This PR Implements

Fixes: https://github.com/getarcaneapp/arcane/issues/2917

## Additional Context

This PR adds two new Trivy scanning options: client/server mode (directing scans at a remote Trivy server rather than downloading the vulnerability DB locally, primarily for arm/v7 hosts) and a flag to suppress unfixable CVEs. The token is correctly passed via `TRIVY_TOKEN` env instead of a CLI flag, and the Zod schema now cross-validates that the server URL is non-empty when server mode is toggled on.

- **Remote server support**: `ScanSourceArgs` replaces local-DB flags with `--server <url>` when a server URL is configured; the auth token is forwarded through the container environment (`TRIVY_TOKEN`) to avoid process-argument exposure.
- **Ignore-unfixed default set to `true`**: The new `TrivyIgnoreUnfixed` setting defaults to `\"true\"` in `DefaultSettingsConfig`, which means all existing installations will silently gain `--ignore-unfixed` on upgrade — scans will stop reporting unfixable CVEs without any user opt-in.
- **Frontend validation**: The environment form schema adds a `superRefine` that surfaces a validation error when server mode is enabled but the URL is blank, catching the misconfiguration at save time.
</details>

Safe to merge after addressing the ignore-unfixed default; the client/server plumbing itself is sound.

The `trivyIgnoreUnfixed` default of `true` in `DefaultSettingsConfig` means every existing installation that upgrades will immediately have `--ignore-unfixed` applied to all future scans. The key won't be present in the database, so it inherits the default on the first settings load. Users who track unfixable CVEs for compliance or backlog purposes will lose those results silently. Everything else — the server URL/token wiring, env-based token handling, and frontend validation — looks correct.

backend/internal/services/settings_service.go — the `TrivyIgnoreUnfixed` default of `"true"` affects all existing deployments on upgrade.
</details>

<sub>Reviews (2): Last reviewed commit: ["feat: allow use of remote trivy server, ..."](https://github.com/getarcaneapp/arcane/commit/66549c79089a919d3d380e29b0c006e8dc043cda) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38842174)</sub>


kmendell:
* **#2999** <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/2999?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/2999?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #3219, opened 2026-07-09
⚡️ Feature: Add Google Chat or generic json payload to webhook notifications


instantchow:
### Feature Description

I would love to see Google Chat setup as a provider of notifications similar to other chat platforms.

### Problem It Solves

There is no built in method getting out a notification via Google Chat.

### Proposed Solution

Either add in a new notification method specifically for Google Chat, which is already something shoutrrr implements via googlechat:// or update the generic webhook to allow for changing the payload template and make variables available for it.

### Alternatives Considered

Creating a gateway service to convert the generic webhook notification to a google chat webhook, would be brittle but I suppose this is the logical alternative that could exist.

### Additional Context

# Straight from shouterr [repo here ](https://github.com/nicholas-fedor/shoutrrr/blob/main/docs/services/chat/googlechat/index.md?plain=1)

# Google Chat

## URL Format

Your Google Chat Incoming Webhook URL will look like this:

!!! info ""
    <https://chat.googleapis.com/v1/spaces/__`FOO`__/messages?key=__`bar`__&token=__`baz`>__

The shoutrrr service URL should look like this:

!!! info ""
    googlechat://chat.googleapis.com/v1/spaces/__`FOO`__/messages?key=__`bar`__&token=__`baz`__

In other words the incoming webhook URL with `https` replaced by `googlechat`.

Google Chat was previously known as Hangouts Chat. Using `hangouts` in
the service URL instead `googlechat` is still supported, although
deprecated.

## Creating an incoming webhook in Google Chat

1. Open the room you would like to add Shoutrrr to and open the chat
room menu.
![Screenshot 1](hangouts-1.png)

2. Then click on *Configure webhooks*.
![Screenshot 2](hangouts-2.png)

3. Name the webhook and save.
![Screenshot 3](hangouts-3.png)

4. Copy the URL.
![Screenshot 4](hangouts-4.png)

5. Format the service URL by replacing `https` with `googlechat`.

### Would you be willing to work on this feature?

- [ ] Yes, I'd like to implement this feature
- [x] I could help with parts of this feature
- [ ] No, I'm just suggesting the feature

kmendell:
Added with https://github.com/getarcaneapp/arcane/commit/20d0f33abb098805e4d4b1713db294801058335e

PR #3417, opened 2026-07-28
feat: custom payload generic webhooks and google chat notifications

khanhx:
## Checklist

- [x] This PR is **not** opened from my fork’s `main` branch
- [x] All new user-facing strings are translated via Paraglide (`m.*()`)

## What This PR Implements

Adds an optional **Payload Template** field to the Generic Webhook notification provider, so users can target endpoints that require a specific request-body shape rather than the flat `{"title": ..., "message": ...}` object.

Addresses: #3219

That issue proposes either a dedicated Google Chat provider *or* "update the generic webhook to allow for changing the payload template and make variables available for it." This implements the latter, which covers Google Chat along with any other endpoint needing a custom envelope (Lark/Feishu, Teams, in-house APIs).

### Why the existing `template` URL parameter doesn't cover this

`BuildGenericURL` already passes a `template` query key through to Shoutrrr, so it looks like this should be configurable today. It isn't:

- Shoutrrr's generic service resolves `template` as the **ID of a template registered on the service instance** (via `SetTemplateString`). `shoutrrr.CreateSenderWithOptions` returns a `*router.ServiceRouter`, which never exposes that service — so there is no seam to register one through the current send path.
- Its built-in `template=json` marshals params into a **flat** JSON object, so nested payloads are unreachable.

Passing a literal body such as `?template={"receiveId":...}` fails with `ErrTemplateNotLoaded`.

Rather than reach into Shoutrrr internals, this renders the template in Arcane and reuses the **existing direct-send path** already built for `SuccessBodyContains`. Providers without a template are untouched and still delegate to Shoutrrr.

## Changes Made

**Backend**

- `models.GenericConfig` gains an optional `template` field.
- `RenderGenericTemplate` renders a Go `text/template` into the request body. Templates receive `title` and `message`, the raw variants `titleRaw` / `messageRaw`, and aliases under the configured `TitleKey` / `MessageKey`.
- `SendGenericWithTitle` routes through the direct-send path when a template is set.
- `buildGenericPayload` centralises body construction; the flat-JSON branch is unchanged.

Two correctness details worth calling out for review:

1. **Values are pre-escaped for JSON string contexts.** Notification text routinely contains quotes and newlines; without escaping, `{"text": "{{.message}}"}` would emit invalid JSON as soon as a message contained a `"`. `titleRaw` / `messageRaw` remain available for non-JSON bodies (e.g. form encoding). `text/template` is used rather than `html/template` deliberately — HTML entity encoding would corrupt webhook payloads.

2. **Rendered output is validated as JSON** when the content type is JSON (or `+json`, or empty, which defaults to JSON). A template can be syntactically valid Go yet still emit malformed JSON — `{"text":"{{.message}}` parses fine — which would otherwise surface only as an unexplained 4xx from the remote endpoint. Validating locally points the user at their own config instead.

The configured-key aliases are only added when they don't shadow a canonical name, so a `TitleKey` of `"message"` cannot repoint `.message` at the title.

**Frontend**

- New `template` field on `GenericFormValues`, wired through `genericSettingsToFormValues` / `genericFormValuesToSettings`.
- Renders as a `textarea` via the existing `DynamicProviderFormBuilder` — no new UI primitive.
- Zod validation rejects a non-empty template with no `{{ . }}` placeholder, which would otherwise send an identical static body for every notification.
- New strings go through Paraglide. The Go-template *example* is a code constant rather than a translatable message: `{{ }}` collides with the message compiler's own placeholder syntax (it breaks `paraglide-js compile`), and the example is code, not prose for translators.

**Tests**

- Backend table tests for rendering, escaping, alias shadowing, JSON validation, and content-type handling.
- `httptest`-backed tests for the full send path: custom template, template + `SuccessBodyContains`, and the invalid-template case asserting the webhook is **not** called.
- A Playwright case covering configuring a template and confirming it survives a reload.

## Testing Done

Ran the development environment locally (frontend + backend) and manually verified end-to-end.

Configured this template in the UI against a local listener:

```json
{"receiveId": "oc_...", "receiveIdType": "chat_id", "msgType": "text", "text": "{{.message}}"}
```

Triggering **Image Update Notification** produced exactly:

```json
{"receiveId": "oc_...", "receiveIdType": "chat_id", "msgType": "text",
 "text": "🔔 Container Image Update Notification\n\nEnvironment: Local Docker\nImage: nginx:latest\n..."}
```

Correct nested envelope, with the multi-line message body correctly escaped as `\n` inside the JSON string.

Also verified:

- The Google Chat shape from the issue (`{"text": "{{.title}}\n{{.message}}"}`) renders and posts correctly.
- The template persists across a full page reload.
- The inline validation error appears for a placeholder-less template.
- A provider with **no** template still sends the original flat `{"title","message"}` payload (backwards compatibility).
- `go test ./pkg/utils/notifications/ ./internal/models/` passes.
- `pnpm -C frontend check` reports 0 errors; `just format` passes for frontend, js, and go.

Not tested: delivery against live Lark/Google Chat tenants, as I don't have accounts on those platforms — verification used local endpoints reproducing their documented payload contracts.

Claude Code (Opus) was used substantially for this change: investigating Shoutrrr's template resolution, drafting the implementation and tests, and driving the browser during manual verification.

Per [AI_POLICY.md](https://github.com/getarcaneapp/arcane/blob/main/AI_POLICY.md), I ran the development environment myself and verified the behaviour described above end-to-end rather than relying on the model's claims. Two issues surfaced during that verification and were fixed rather than papered over: the escaping requirement for quotes/newlines in JSON string contexts, and the alias-shadowing bug when `TitleKey` collides with a canonical variable name. I reviewed and edited this description.

## Additional Context

Fully backwards compatible — the field is optional, and an empty template preserves the current flat-JSON behaviour through Shoutrrr.

Happy to adjust the design if maintainers would prefer a dedicated `googlechat://` provider instead, or in addition. The two aren't mutually exclusive: a first-class provider is a nicer experience for that specific platform, while the template covers the long tail of custom endpoints.

One open question for reviewers: the template renders with `missingkey=zero`, so an unknown variable yields an empty string rather than erroring. I chose leniency to avoid breaking a working notification at send time, but I'm happy to make it strict if you'd rather surface typos loudly.

Adds customizable generic-webhook payload templates and a dedicated Google Chat notification provider.
- Exposes event metadata to payload templates and validates rendered JSON when settings are saved.
- Adds notification settings UI, shared types, translations, and end-to-end coverage for the new providers.
- Extends generic delivery with template registration, direct response-body validation, and bounded HTTP clients.
</details>

The PR does not appear safe to merge until templated generic webhooks consistently honor inline URL options and configured payload templates.

Response-body validation still switches to a direct sender that does not interpret Shoutrrr configuration encoded in the webhook URL, and template=json or template=JSON still bypasses a saved custom payload template in the ordinary send path.

**Files Needing Attention:** backend/pkg/utils/notifications/generic_sender.go
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/generic-we..."](https://github.com/getarcaneapp/arcane/commit/7b1b38d8072499d0a9f51e07ab23d1404fad0320) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47811182)</sub>


kmendell:
Ill try to finish this up as soon as i can, over all looked fine, just need to make sure my stuff didnt break it and some cleanup

kmendell (approved):
LGTM

PR #3554, opened 2026-08-09
fix: editor line highlight and selection rendering fully opaque on default accent color

### LINKED ISSUE #3492: 🐞 Bug: Editor Line Highlight/Selection render fully opaque on accounts that have not customized accent color
### Bug Description

On any account that has never explicitly picked an accent color — the out-of-box state of every account — the project editor's active-line highlight and selection render at **full opacity** in the accent purple instead of the intended 5% / 35% alpha, and the caret, drawn in that same accent, is invisible while it's on the highlighted line.

Picking any other accent color in Account settings fixes the account **permanently** — even switching back to Default afterward stays healthy, across reloads and re-logins. On a fixed account, the only way to bring the breakage back is to force it through the custom accent color picker with a non-`oklch(...)` color pattern — e.g. `rgb(139, 92, 246)`, which is the default purple in a different notation, breaks it deterministically.

**The cause, in plain terms:** the editor theme applies transparency to the accent color in JavaScript, but that code only understands colors written as `oklch(...)` or 6-digit `#hex`. The production CSS bundle actually ships the default accent as `lab(...)` (the build tooling rewrites it), so the transparency function returns it unchanged — fully opaque. Explicitly picking an accent stores an `oklch(...)` string the code *does* understand, which is why the bug permanently vanishes for that account.

Full code-level analysis (verified against the shipped v2.6.0 CSS bundle and the `v2.6.0` tag), the other input formats affected, and a suggested one-line fix are in **Additional Context** at the bottom. This also looks like the probable root cause behind #3468 — the caret isn't independently broken; it's just invisible against a solid line highlight in its own color.

### Steps To Reproduce

1. On any Arcane instance, create a **new user** and log in as that user.
2. Do **not** change any appearance settings.
3. Open any project's editor, or create a new project (either works).
4. Observe: the active line renders with a solid, full-opacity purple bar (text still legible against it, but low-contrast); the caret is invisible while on the highlighted line. Selecting text on the active line *appears to do nothing* — selection is the same accent at nominally higher alpha, but both render fully opaque, so selection and line highlight are indistinguishable. Drag across multiple lines and the selection shows up as the same solid purple on the other lines.
5. Now click your user (bottom left) → **Account**, scroll down to the accent colors, and pick any color **other than Default**. (Clicking Default does nothing — the radio is already on Default, so no change event fires.)
6. Return to the editor: highlight and selection now render at the intended subtle alpha, caret visible.
7. The fix is permanent for that account — you can even switch back to Default now and it stays correct, surviving reloads and re-login. (See Additional Context: an explicitly clicked Default stores a real color value; the out-of-box state stores nothing.)

**Additional reproduction method (confirmed):** on a healthy account, open the custom accent color dialog and enter `rgb(139, 92, 246)` — the default purple, just written in a different notation. The solid opaque bar comes back immediately. `hsl(...)` and named colors (e.g. `gold`) behave the same way.

### Expected Behavior

The editor's active-line highlight renders as a subtle ~5% accent tint, selection at ~35% is clearly distinguishable from the line highlight, and the caret is visible at all times — i.e. exactly what the editor renders on the same account immediately after any accent color has been explicitly chosen (see the "Change to Amber" screenshot).

### Actual Behavior

On an account that has never explicitly picked an accent color: the active line is a solid, fully opaque accent-purple bar (poor text contrast); the caret is invisible whenever it's on that line; selection renders at the same full opacity, making it indistinguishable from the line highlight on the active line. The nominal 5% / 15% / 35% alpha tiers all collapse into the same solid color — consistent with the alpha-derivation function returning its input verbatim.

### Screenshots

Same project, same account, dark mode:

**First login (broken):** fresh account, untouched settings — solid purple active-line highlight in both panes; the caret is on the `.env` line of the Compose pane and is completely invisible.

Image

**After picking Amber:** correct subtle line tint, caret clearly visible.

Image

**After switching from Amber back to Default:** still correct, and stays correct across reloads/re-logins. Same nominal color as the broken first shot — the difference is only *how* the value reaches the editor (see Additional Context).

Image

### Arcane Version

v2.6.0

### Installation Method

Docker Compose (Recommended)

### Environment Type

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

Docker version 29.7.1, build e9452d6

### Browser & Version

Chrome 151.0.7922.71 (Official Build) (x86_64) - via AUR

```shell

```

```yaml
services:
  arcane:
    image: ghcr.io/getarcaneapp/manager:latest
    container_name: arcane
    ports:
      - "3552:3552"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - arcane-data:/app/data
      - /etc/arcane/projects:/etc/arcane/projects
      - /host/path/to/builds:/builds
    environment:
      - ENCRYPTION_KEY=<redacted>
      - JWT_SECRET=<redacted>
      - PROJECTS_DIRECTORY=/etc/arcane/projects
      # Optional: run Arcane as a specific host UID/GID instead of the default runtime user.
      # When using the mounted Docker socket, Arcane will map the socket group automatically.
      # - PUID=1000
      # - PGID=1000
      # Timezone for cron job scheduling (e.g., America/New_York, Europe/London, UTC)
      # See: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
      - TZ=UTC
    # Use host cgroup namespace for better container self-detection
    # This allows Arcane to detect its own container ID more reliably
    cgroup: host
    healthcheck:
      test:
        ["CMD", "./arcane", "health", "--timeout", "2s"]
      interval: 10s
      timeout: 3s
      retries: 5
      start_period: 15s
    restart: unless-stopped

volumes:
  arcane-data:
```

```shell
# Arcane Environment Configuration
# Copy this file to .env and customize the values for your production setup
# Generate secure values with: openssl rand -base64 32

# Application Settings
GIN_MODE=release
ENVIRONMENT=production
PORT=3552
APP_URL=http://localhost:3552

# Optional direct TLS server mode (HTTP/2 over TLS)
# Leave disabled when running behind a reverse proxy that terminates TLS.
# TLS_ENABLED=false
# TLS_CERT_FILE=/path/to/fullchain.pem
# TLS_KEY_FILE=/path/to/privkey.pem

# When running behind a reverse proxy (Caddy, Nginx, Traefik), set the CIDR
# range(s) of the proxy so X-Forwarded-* headers are trusted for client IP
# and TLS scheme detection. Auto-defaults to loopback ranges when LISTEN is
# bound to 127.0.0.1 or ::1. Comma-separated, e.g.:
# TRUSTED_PROXIES=127.0.0.0/8,::1/128

# Optional runtime UID/GID for published Docker images.
# When both are set, Arcane will run as that UID/GID and files it writes under
# /app/data will be owned by that user inside the container.
# Leave both unset to keep the current default runtime behavior.
# PUID=1000
# PGID=1000

# File and Directory Permissions (Octal)
# Default: 0644 for files, 0755 for directories
# FILE_PERM=0644
# DIR_PERM=0755

# Security Settings
# IMPORTANT: Generate unique values for production!
#ENCRYPTION_KEY=your-32-char-encryption-key-here
#JWT_SECRET=your-super-secret-jwt-key-change-this
# Optional: bootstrap a deterministic admin API key for IaC workflows (supports _FILE / __FILE variants)
# ADMIN_STATIC_API_KEY=arc_your_deterministic_admin_api_key
# JWT refresh token expiry duration (default: 168h = 7 days). Accepts Go duration format (e.g., 72h, 720h, 8760h).
# JWT_REFRESH_EXPIRY=168h

# Database Configuration
DATABASE_URL=file:data/arcane.db?_pragma=journal_mode(WAL)&_pragma=busy_timeout(2500)&_txlock=immediate
# Downgrades are blocked by default; enable only temporarily after taking a backup.
ALLOW_DOWNGRADE=false
# Optional: GitHub API token used for downgrade migration fetches to avoid rate limits.
# GITHUB_TOKEN=ghp_your_token_here

# Docker Configuration
# DOCKER_HOST=unix:///var/run/docker.sock  # Default: direct socket access
# DOCKER_HOST=tcp://docker-socket-proxy:2375  # Example: via socket proxy for enhanced security
# Trivy scan container security options (optional, useful on SELinux hosts)
# Separate multiple entries with commas or new lines, for example: label=disable,label=type:container_runtime_t
# TRIVY_SECURITY_OPTS=label=disable
# Run Trivy scan containers in privileged mode only when security options alone are not enough
# TRIVY_PRIVILEGED=false

# Agent Configuration (optional)
# AGENT_MODE=true
# AGENT_TOKEN=api-token

# ============================================================================
# AUTO-LOGIN CONFIGURATION - FOR DEVELOPMENT/TESTING ONLY
# ============================================================================
# WARNING: Do NOT enable auto-login in production environments!
# This feature automatically logs in users without requiring credentials,
# which is a significant security risk if exposed publicly.
#
# Use cases:
# - Local development environments
# - Automated testing (e.g., Playwright E2E tests)
# - Demo environments with restricted network access
#
# When enabled, the frontend will automatically authenticate using the
# configured credentials on application startup (once per session).
# ============================================================================
# AUTO_LOGIN_ENABLE=false
# AUTO_LOGIN_USERNAME=arcane
# AUTO_LOGIN_PASSWORD=arcane-admin
```

### Additional Context

Actual install method is to a Debian 13 LXC running in Proxmox. Installed via Docker LXC from PVE Helper-Scripts, with Arcane installled via Addon script from the same source.

**Methodology disclosure:** I'm an experienced infrastructure/network engineer, not a developer - I found and reproduced the problem, but the code analysis below was done via Claude Fable 5 at my direction, against the `v2.6.0` tag, and then verified empirically against my live instance. Since I know you folks use Claude in your own workflow, I've asked Claude to make this self-contained enough that you can hand it straight to your tooling and validate it without needing follow-ups from me. I am however available for followups as needed.

## Root cause

`frontend/src/lib/components/code-editor/theme.ts` derives editor colors from the accent:

```ts
caret: accentColor,
selection: accentWithAlpha35,
selectionMatch: accentWithAlpha15,
lineHighlight: accentWithAlpha05,
```

where the accent is read live from CSS: `getComputedStyle(document.documentElement).getPropertyValue('--primary')`, and alpha is applied by `getAccentColorWithAlpha()`, which handles exactly **two** input formats:

- strings starting with `oklch` → alpha appended inside the function
- strings starting with `#` → parsed as 6-digit hex into `rgba()`

**Any other format falls through and is returned verbatim — i.e. at full opacity.**

The trap: the *source* stylesheet defines `--primary: oklch(0.606 0.25 292.717)` (and `oklch(0.541 0.281 293.009)` for `.dark`) — which would be handled. But the **production build compiles those values to `lab()`**: LightningCSS lowers out-of-sRGB-gamut oklch colors to `lab()` for gamut preservation, and the default violet-500/600 accents are out of gamut. From the shipped `v2.6.0` CSS bundle (`_app/immutable/assets/0.*.css`):

```
:root { … --primary:lab(49.9355% 55.1776 -81.8963) … }   /* violet-500 */
.dark { … --primary:lab(41.088% 68.9966 -91.995) … }     /* violet-600 */
```

(No `@property` registration for `--primary` exists in the bundle, so `getComputedStyle` returns this token stream verbatim.)

Failure chain for an out-of-box account:

1. No stored accent preference → `applyAccentColor('default')` → the inline `--primary` override is **removed** → the value cascades from the **built stylesheet** → `lab(41.088% 68.9966 -91.995)`.
2. `getAccentColorWithAlpha()` doesn't recognize `lab(` → returns it unchanged → `lineHighlight`, `selection`, and `selectionMatch` are all **fully opaque** accent purple.
3. `caret: accentColor` is that same purple → a caret blinking against a solid same-color line highlight → invisible caret (the #3468 symptom).

Why picking a color heals it — and why Default-clicking is a no-op:

4. The accent picker's swatches carry raw `oklch(…)` strings as their values (including the Default swatch: `'oklch(0.606 0.25 292.717)'`). Picking one stores that string as the user's preference and applies it **inline** via `document.documentElement.style.setProperty('--primary', …)` — JavaScript, so the build transform never touches it. `getComputedStyle` now returns `oklch(…)` → handled branch → correct alpha, permanently (the stored string is re-applied inline on every login).
5. On an out-of-box account, the picker *renders* the Default swatch as already selected, so clicking it fires no change event and nothing is stored — the broken state can't be exited via the Default swatch. Selecting a different color and then returning to Default works, because at that point Default is an actual value *change* and the raw `oklch(…)` string gets stored like any other pick. There are effectively two states the UI renders identically as "Default": *no preference stored* (broken) and *Default explicitly chosen after another color* (healthy).

Notably, the in-sRGB-gamut theme accents (e.g. the Nord palette) compile to plain 6-digit hex — the handled branch — so switching application themes can also mask the bug.

## Related inputs that hit the same fallthrough

- The **custom accent color dialog** is a free-text input validated only by "does the browser parse it". `rgb(139, 92, 246)` reproduces the full-opacity breakage deterministically (confirmed on v2.6.0). `hsl(…)` and named colors (`gold`, `rebeccapurple`) likewise. The backend validator (`types/settings/accent_color.go`, `SafeAccentColor`) explicitly *allows* named colors, `rgb()`, `hsl()`, and `oklab()` — formats the frontend alpha derivation can't process.
- **3- and 4-digit hex** (`#abc`, `#abcd`): the fixed-position slicing produces `rgba(NaN, NaN, NaN, α)`, an invalid color, so the browser drops the declaration — selection and line highlight disappear entirely (an *invisible*-selection variant of the same bug).
- A fully transparent custom accent (`#00000000`) makes the caret and selection vanish (caret inherits the accent's alpha).

## Suggested fix

Replace the format-sniffing in `getAccentColorWithAlpha()` with CSS-native alpha compositing, which is format-proof for every color the browser can represent (`lab()`, `oklch()`, any-length hex, `rgb()`, `hsl()`, named):

```ts
function getAccentColorWithAlpha(alpha: number): string {
	return `color-mix(in srgb, ${getAccentColor()} ${alpha * 100}%, transparent)`;
}
```

`color-mix()` is supported by every browser that can run Arcane, and the codebase already uses it for `--ring` in `applyAccentColor()`.

Optional hardening, independent of the above: drawing the caret in `palette.foreground` instead of the accent (and/or widening `.cm-cursor` slightly from the default ~1.25px) would keep the caret visible even under pathological accent choices — but with the alpha fix in place the accent caret is perfectly usable, as the "Back to Default" screenshot shows.

## Maintainer-side verification (no reporter round-trip needed)

- On any account still in its out-of-box accent state, run in devtools: `getComputedStyle(document.documentElement).getPropertyValue('--primary')` → returns `lab(41.088% 68.9966 -91.995)` (dark mode) instead of an `oklch(…)` string.
- Inspect the shipped bundle: `grep -oE -- '--primary:[^;]+' 0.*.css` inside the image's frontend assets → values are `lab()` / hex, never `oklch()`.
- Code path: `frontend/src/lib/components/code-editor/theme.ts` (`getAccentColor` / `getAccentColorWithAlpha`) vs. `frontend/src/lib/utils/theme.ts` (`applyAccentColor`) and `frontend/src/lib/components/accent-color/accent-color-picker.svelte` — identical on `v2.6.0` and current `main`.

kmendell:
## Checklist

## What This PR Implements

Fixes: https://github.com/getarcaneapp/arcane/issues/3492

## Additional Context

This change replaces format-specific CodeMirror accent alpha handling with CSS `color-mix()` so editor selections, match highlights, and active-line backgrounds work with any supported CSS color notation.

A real Chromium CodeMirror rendering check compared the prior and current behavior with `oklch`, hexadecimal, `rgb`, `hsl`, and `lab` primary colors. The concern that non-OKLCH and non-hex colors would remain opaque was disproved: the current implementation generated transparent overlay colors at the intended 35%, 15%, and 5% opacity levels for every tested format. No defects were found.
</details>

Safe to merge: the editor color-overlay behavior was rendered and verified in Chromium across representative supported CSS color formats.

The focused browser check exercised the changed CodeMirror theme behavior and compared it with the previous implementation. It confirmed that formats previously left opaque now resolve to transparent selection, match, and active-line backgrounds.

**Files Needing Attention:** No files need follow-up attention.
</details>

<details><summary><h3><a href="https://www.greptile.com/trex">T-Rex</a> T-Rex Logs</h3></summary>

**What T-Rex did**
- Ran the CodeMirror primary-color browser validation script and it completed with exit code 0.
- Rendered a real CodeMirror editor in Chromium with reconstructed theme initializers using oklch, hex, rgb, hsl, and lab primary colors.
- Compared the before and after color handling: earlier results produced opaque rgb/hsl/lab values, while the updated implementation used color-mix with transparent and Chromium computed 35%, 15%, and 5% alpha overlays.
- Viewed the structured CSS and computed-color results artifact to verify alpha handling across formats.
- Watched the after-rendering video artifact to confirm the updated CodeMirror rendering preserves overlay transparency.

<a href="https://app.greptile.com/trex/runs/18011921/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

<sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: editor line highlight and selection..."](https://github.com/getarcaneapp/arcane/commit/3ba508409a412760a58b63d79c6a33b4fbac6626) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51600845)</sub>

### DIFF
--- a/frontend/src/lib/components/code-editor/theme.ts
+++ b/frontend/src/lib/components/code-editor/theme.ts
@@ -8,24 +8,10 @@ function getAccentColor(): string {
 }
 
 function getAccentColorWithAlpha(alpha: number): string {
-	const accentColor = getAccentColor();
-
-	if (accentColor.startsWith('oklch')) {
-		const hasAlpha = accentColor.includes('/');
-		if (hasAlpha) {
-			return accentColor.replace(/\/\s*[\d.]+\s*\)/, ` / ${alpha})`);
-		}
-		return accentColor.replace(')', ` / ${alpha})`);
-	}
-
-	if (accentColor.startsWith('#')) {
-		const r = parseInt(accentColor.slice(1, 3), 16);
-		const g = parseInt(accentColor.slice(3, 5), 16);
-		const b = parseInt(accentColor.slice(5, 7), 16);
-		return `rgba(${r}, ${g}, ${b}, ${alpha})`;
-	}
-
-	return accentColor;
+	// color-mix handles any color notation the CSS pipeline may emit for
+	// --primary (oklch, lab, rgb, hex, ...) — format-sniffing here previously
+	// returned unknown formats unchanged, i.e. fully opaque.
+	return `color-mix(in oklab, ${getAccentColor()} ${Math.round(alpha * 100)}%, transparent)`;
 }
 
 interface ArcaneThemePalette {


kmendell:
* **#3554** <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3554?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #862, opened 2025-11-07
🐞 Bug: toomanyrequests You have reached your unauthenticated pull rate limit


banghasan:
### Bug Description

Could you please help me understand why this is happening?
I’ve already logged into Docker via the console and successfully pulled the image, but I’m encountering this error in Arcane.

### Steps To Reproduce

1. create projects
2. compose
3. up

### Expected Behavior

I hope it can run normally and pull images without any issues.

### Actual Behavior

```
failed to pull image adminer: failed to initiate image pull for adminer: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

### Screenshots

![Image](https://github.com/user-attachments/assets/7aee8634-9c85-46b1-8de6-a8384cc88d38)

### Arcane Version

v1.7.2

### Operating System

Linux

### Browser

Firefox 144.0.2 

### Docker Version

 28.5.2

kmendell:
Do you have docker credentials added in arcane?

banghasan:
> Do you have docker credentials added in arcane?

How can I do that?

banghasan:
I’ve found it! I completely forgot about that. Thank you so much!

piejanssens:
For others like me that ended up here after a search: https://getarcane.app/docs/features/images#private-registry-authentication

PR #3639, opened 2026-08-17
fix(registry): use stored credentials for non-Docker Hub registries

### LINKED ISSUE #2635: 🐞 Bug: Update Check Failed: Error response from daemon: toomanyrequests
### Bug Description

At any given time, several of my projects will have a red exclamation mark icon in the "Updates" column, and show the following error (this example is for radarr, but they're all the same 'toomanyrequests' error).

```
Update Check Failed
distribution inspect failed for ghcr.io/linuxserver/radarr:latest: Error response from daemon: toomanyrequests: retry-after: 511.05µs, allowed: 44000/minute
```

### Steps To Reproduce

Nothing, arcane just isn't displaying whether there's an update available because there's a red icon with the 'toomanyrequests' error.

### Expected Behavior

Projects should almost always have either a green "Up to Date" icon or a blue "Has Updates" icon.

### Actual Behavior

Nothing, arcane just isn't displaying whether there's an update available because there's a red icon with the 'toomanyrequests' error. If I manually click on the red icon and check for an update, it works.

### Screenshots

_No response_

### Arcane Version

v1.19.2

### Installation Method

Docker Compose (Recommended)

### Environment Type

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

28.3.3

### Browser & Version

Zen 1.19.13b (Firefox 150.0.3) 

```shell

```

```yaml

```

```shell

```

### Additional Context

_No response_

BobzTH:
Fixes #2635.

## The problem

`inspectImageDigestViaDaemonInternal` only applies stored credentials up front when the registry is Docker Hub:

```go
if isDockerHubRegistryInternal(registryHost) {
    credentials, credErr := s.getMatchingRegistryCredentialsInternal(ctx, registryHost, externalCreds)
    if credErr == nil && len(credentials) > 0 {
        return s.inspectImageDigestWithCredentialsInternal(...)
    }
}

inspectResult, err := dockerClient.DistributionInspect(ctx, normalizedRef, client.DistributionInspectOptions{}) // anonymous
if !isUnauthorizedRegistryErrorInternal(err) {
    return ..., errors.WrapIff(err, "distribution inspect failed for %s", normalizedRef)
}
```

Every other registry goes anonymous first, and credentials are only retried when the anonymous attempt fails as *unauthorized*. `toomanyrequests` is a rate limit, not an unauthorized error, so it returns at the guard above and the configured credential is never reached:

```
ghcr.io/getarcaneapp/agent:latest: distribution inspect failed for ghcr.io/getarcaneapp/agent:latest:
Error response from daemon: toomanyrequests: retry-after: 1.246809ms, allowed: 44000/minute
```

`allowed: 44000/minute` is GHCR's shared unauthenticated quota, so this fires at random based on global traffic. The retry in `InspectImageDigest` re-attempts anonymously up to 5 times, so it cannot recover either. From the user's side a ghcr.io credential is configured and enabled but has no effect, which is what #2635, #862 and the follow-up reports describe. It shows up most on the Arcane images themselves because every install polls them.

## The change

Apply matching stored credentials up front for any registry, not just Docker Hub.

Docker Hub keeps its existing fail-fast behaviour, since its anonymous quota is too small for an anonymous retry to be worth attempting. For other registries a failed credentialed attempt now falls back to anonymous, so a stale or under-scoped token cannot break public images that resolve fine without one.

Removing the Docker Hub gate makes `getMatchingRegistryCredentialsInternal` reachable for every registry, including when the service holds no database, so it now returns no credentials there instead of panicking inside `GetEnabledRegistries`. That path is already exercised by the existing anonymous tests, which construct the service with a nil DB.

## Tests

Three tests added to `backend/internal/registry/service_test.go`:

- `..._UsesStoredCredentialsForNonDockerHubRegistry` — a stored ghcr.io credential is sent on the first inspect
- `..._UsesStoredCredentialsInsteadOfRateLimitedAnonymous` — reproduces #2635 exactly; the fake daemon returns the `toomanyrequests` string above for anonymous calls and succeeds when authenticated
- `..._FallsBackToAnonymousWhenStoredCredentialRejected` — a rejected credential still resolves a public image anonymously

The first two fail on `main` and pass with this change. Verified `go test ./internal/registry/` is green, and that the four failures in a full `go test ./...` run (`frontend`, `internal/project` ×2, `pkg/libarcane`) reproduce identically on unpatched `main` in the same environment, so they are unrelated to this change.

## Verified on real hardware

Built this branch with `docker/Dockerfile` and ran it against a live reproduction — an Arcane manager on a Synology NAS with four edge agents, with both a ghcr.io and a Docker Hub credential configured.

`image_updates` before, on v2.8.0:

| repository | auth_method | used_credential |
|---|---|---|
| ghcr.io/immich-app/immich-server | anonymous | 0 |
| ghcr.io/getarcaneapp/manager | anonymous | 0 |
| germannewsmaker/myspeed | credential | 1 |

After, on this branch:

| auth_registry | auth_method | images | with_cred |
|---|---|---|---|
| docker.io | credential | 10 | 10 |
| ghcr.io | credential | 4 | 4 |

Scan reported `checked=15 updates=0 errors=0`, container healthy, all four edge agents reconnected, and Docker Hub behaviour unchanged.

## AI disclosure

Per AI_POLICY.md: written with Claude Code. It located the root cause, wrote the change and the tests, and ran the builds and test runs. I directed the investigation from a live failure on my own hardware, and the before/after verification above is from that deployment rather than a hypothetical. Happy to adjust the Docker Hub asymmetry or the anonymous fallback if you'd prefer different semantics there.


kmendell (approved):

kmendell:
LGTM, Thanks!

jdrouhard:
Hello! Thanks for attempting to fix this! Do you have any ideas as to why I'm still getting this error with 2.8.1 (which contains your fix)?

I regularly have one image check that will fail, and it's always lscr (the specific image varies):

> lscr.io/linuxserver/plex:latest: distribution inspect failed for lscr.io/linuxserver/plex:latest: Error response from daemon: toomanyrequests: retry-after: 835.359µs, allowed: 44000/minute

BingoRox:
> > lscr.io/linuxserver/plex:latest: distribution inspect failed for lscr.io/linuxserver/plex:latest: Error response from daemon: toomanyrequests: retry-after: 835.359µs, allowed: 44000/minute

@jdrouhard the GitHub PAT authentication doesn't work with `lscr.io` images. Linux Server created this as a pseudo-registry for telemetry and load balancing. It uses ghcr.io as it's backend registry so basically just change any image that starts with `lscr.io` to `ghcr.io` and authentication will work, allowing the fix implemented in this PR to also work. 

`lscr.io/linuxserver/plex:latest` > `ghcr.io/linuxserver/plex:latest`

More info about lscr.io: https://www.linuxserver.io/blog/wrap-up-warm-for-the-winter

kennethsible:
@BingoRox I just wanted to mention that I'm also still experiencing these errors daily, even after this PR was merged. I've already replaced `lscr.io` with `ghcr.io` in my compose file and repulled the images since I figured that was the reason. I can also confirm that Arcane shows my auth credentials for GitHub when I hover over the update icon for `ghcr.io` images.

jdrouhard:
I've dug into this quite a bit.

lscr.io does use GitHub PAT tokens--all auth requests are forwarded to GitHub. There used to be some code that mapped lscr.io internally to ghcr.io so they shared concurrency semaphores and rate limits, but it was removed in a refactor. So now they are considered separate registries, and having a mix from each causes arcane to exceed the built in defaults.

There's also an issue with the fix in this PR. Non-auth errors from an authenticated request (like rate limits) still fallback to an anonymous attempt, actually causing _more_ requests under rate limiting than less. ghcr (at least) doesn't change the 44000/minute rate limit for an authenticated request so having an API key doesn't actually decrease the frequency these 429s are returned. And the http registry (vs daemon) fallback function didn't get the same authorized-first change.

I've got some changes I'm going to push to a PR soon. I'm testing locally at the moment. However--even fixing all of the above--it's still possible to hit limits when these checks are occurring right on the hour. The real fix is to parse the retry-after header and actually use it.

I'm also experimenting with moving the cron job to run at a random minute within the hour which I think will help avoid the mass rush of api requests from everyone's on-the-hour update checks everywhere.

PR #3682, opened 2026-08-20
fix(api): percent-decode path parameters before use

### LINKED ISSUE #3657: 🐞 Bug: API does not percent-decode `{imageId}` path segments; encoded IDs silently return empty vulnerability lists
### Bug Description

The API extracts `{imageId}` path parameters without percent-decoding them and passes the raw
string straight to Docker / the database. A client that percent-encodes the ID — which RFC 3986
permits, and which some HTTP clients and proxies do by default — gets wrong answers.

The clearest proof uses an **unreserved** character. In `sha256:08e4…`, the leading `s` is
unreserved (RFC 3986 §2.3), and §6.2.2.2 states that two URIs differing only in the
percent-encoding of unreserved characters **are equivalent**. So `%73ha256:…` and `sha256:…`
are the same URI and must behave identically. They do not:

```
GET /api/environments/0/image-updates/check/sha256:9146…    -> 200 OK
GET /api/environments/0/image-updates/check/%73ha256:9146…  -> 500
```

and the error body echoes the undecoded string back:

```json
{"title":"Internal Server Error","status":500,
 "detail":"Failed to check image update: failed to get image reference: image not found:
           no local image or running container found for %73ha256:9146935e0797ce14a340616ceab2f77b3bc61d1b10befb0718a4be7f01485e2c"}
```

`%73ha256:` is not a Docker reference — it is the literal request path. Nothing decoded it.

**How this surfaced.** We first hit it encoding the `:` as `%3A`, which is the natural thing for
a client to do since `:` is a *reserved* character (§2.2):

```
GET /api/environments/0/image-updates/check/sha256%3A9146…  -> 500
   detail: "… no local image or running container found for sha256%3A9146935e…"
```

That case alone is arguable — one could say `:` is legal unencoded in a path segment, so don't
encode it. The `%73` case removes that defence: for unreserved characters the two forms are the
same URI by specification, and the server must decode.

### The failure mode that worries us most: HTTP 200 with zero items

On the paginated endpoint the request does not fail at all. It succeeds and reports **no
vulnerabilities** for an image that has 44:

```
GET /api/environments/0/images/sha256:08e4…/vulnerabilities/list?limit=2
-> 200  "pagination":{"totalItems":44,"grandTotalItems":44, ...}

GET /api/environments/0/images/sha256%3A08e4…/vulnerabilities/list?limit=2
-> 200  {"success":true,"data":[],
         "pagination":{"totalPages":1,"totalItems":0,"currentPage":1,"itemsPerPage":2,"grandTotalItems":0}}

GET /api/environments/0/images/%73ha256:08e4…/vulnerabilities/list?limit=2
-> 200  {"success":true,"data":[], "pagination":{ … "grandTotalItems":0}}
```

Same image (`curlimages/curl:8.5.0`, scan status `completed`, 6 HIGH / 21 MEDIUM / 17 LOW),
`success: true` in every case. A client that encodes its path parameter concludes the image is
clean. There is no error to log, no status code to alert on, and the pagination block actively
asserts `grandTotalItems: 0`. This is a silent false negative on security data.

The cause is visible in `ListImageVulnerabilities`
(`backend/internal/vulnerability/handler.go:343`): `input.ImageID` is forwarded verbatim into
`vulnerabilityService.ListVulnerabilities`, where it becomes a query filter. A filter that
matches nothing is not an error, so the handler normalises `nil` to `[]` and returns 200.

The behaviour is inconsistent across endpoints for the same bad input, which makes it hard for a
client to detect:

| Endpoint | encoded ID |
|---|---|
| `GET …/image-updates/check/{imageId}` | 500 |
| `GET …/images/{imageId}/vulnerabilities` | 404 `Vulnerability scan not found` |
| `GET …/images/{imageId}/vulnerabilities/list` | **200, empty, `success: true`** |

### Steps To Reproduce

```bash
B=http://<host>:3552/api
KEY=<api-key>
# any image id from GET $B/environments/0/images
ID=sha256:9146935e0797ce14a340616ceab2f77b3bc61d1b10befb0718a4be7f01485e2c

curl -s -o /dev/null -w "raw  %{http_code}\n" -H "X-API-Key: $KEY" "$B/environments/0/image-updates/check/$ID"
curl -s -o /dev/null -w "%3A  %{http_code}\n" -H "X-API-Key: $KEY" "$B/environments/0/image-updates/check/${ID/:/%3A}"
curl -s -o /dev/null -w "%73  %{http_code}\n" -H "X-API-Key: $KEY" "$B/environments/0/image-updates/check/%73${ID#s}"
```

For the silent case, substitute an image that has a completed vulnerability scan and call
`…/images/{imageId}/vulnerabilities/list` with the three forms, comparing
`pagination.grandTotalItems`.

### Expected Behavior

`{imageId}` is percent-decoded before use, so `sha256:…`, `sha256%3A…` and `%73ha256:…` all
resolve to the same image.

### Actual Behavior

The raw path segment is passed through to Docker and to the vulnerability store. Depending on the
endpoint this yields 500, 404, or a successful-looking empty result.

### Suggested Fix

The codebase already has the right helper and applies it to the sibling `{name}` parameter —
`decodeImageNameInternal` at `backend/internal/image/handler.go:538` wraps `url.PathUnescape`,
and `GetImageAttestations` uses it at line 411. The `{imageId}` inputs
(`backend/internal/image/handler.go:69` and `:123`, and the four in
`backend/internal/vulnerability/handler.go:27,36,45,63`) never go through it, so this looks like
an oversight rather than a deliberate contract. Applying the same decode to `{imageId}` — or
decoding once in the routing layer so every handler is covered — would fix all of the above.

Separately, it may be worth having the vulnerability list distinguish "no scan for this image ID"
from "scan exists, zero vulnerabilities", so an unresolvable ID can never be reported as a clean
image.

### Arcane Version

v2.8.0. Verified still present in v2.8.1: the `v2.8.0...v2.8.1` diff contains no
`PathUnescape` / `QueryUnescape` / `RawPath` changes, and `backend/internal/image/handler.go` is
byte-identical between `v2.8.1` and `main` at the time of writing.

### Installation Method

Docker Compose (Recommended)

### Environment Type

kmendell:
## Checklist

## What This PR Implements

Fixes: https://github.com/getarcaneapp/arcane/issues/3657

## Additional Context

This change centralizes percent-decoding for route parameters, removes duplicate decoding from image and template handlers, and changes a missing vulnerability scan list response from an empty successful page to a not-found response.

Escaped image and template path segments were exercised before and after the router change. Encoded slashes continued to select the same routes, while encoded colons and double-encoded percent characters were delivered to handlers after exactly one decode. This disproved the route-selection concern.

The missing-scan backend path was exercised against the real persistence-backed service and Huma handler: it now produces the intended not-found response instead of an empty successful page.

### T-Rex validation blocked
The focused frontend check for handling the new missing-scan response could not run because the installed frontend test setup could not resolve the Vitest package. The exact missing item was the Vitest package required by that test configuration; no frontend runtime result was obtained.
</details>

No confirmed defect remains; the changed router behavior preserves route selection while supplying once-decoded parameter values.

Executable checks covered route matching and path-value behavior before and after the change, as well as the persistence-backed missing-scan backend response. No verified finding was produced.

**Files Needing Attention:** No files require changes from this review. The frontend vulnerability list consumer would benefit from a runnable check covering a not-found list response once the frontend test dependency is available.
</details>

<details><summary><h3><a href="https://www.greptile.com/trex">T-Rex</a> T-Rex Logs</h3></summary>

**What T-Rex did**
- T-Rex completed the requested contract validations for proofs 0 through 3, but local artifact references were not uploaded.

<sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(api): percent-decode path parameters..."](https://github.com/getarcaneapp/arcane/commit/0d82182fe5e49cdd9940c62f140dc8004ea2ed36) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55029476)</sub>


kmendell:
* **#3682** <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3682?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3682?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3742, opened 2026-08-25
fix: rate-limit webhook triggers by token instead of client IP

### LINKED ISSUE #3652: 🐞 Bug: Webhook rate limiting causes failures with multiple simultaneous Git-sync projects from a single push
### Bug Description

A single push to the monorepo that touches multiple project subfolders (or even a push that Forgejo's webhook delivery fires for all 14 configured webhooks near- simultaneously, regardless of which paths actually changed) results in some of the webhook-triggered syncs failing with:

{"error":"rate limit exceeded"}

Observed via raw HTTP response headers from one of the failed requests:

HTTP/2 429 (implied by Retry-After)
Retry-After: 60
Content-Type: application/json
Content-Length: 31
Server: cloudflare

### Steps To Reproduce

1.Configure 10+ separate Arcane Git-sync projects, each pointing to a different subfolder of the same monorepo, each with its own individual webhook configured in Git (Forgejofor me) (Webhook, Push Events, pointing to that project's Arcane webhook URL).
2.Push a commit to main (in my case, a single commit touching files across the repo, causing Forgejo to fire all 10+ webhooks in quick succession).
3.Observe: the first ~10 webhook deliveries succeed, the remaining fail with {"error":"rate limit exceeded"}.

### Expected Behavior

The rate limit threshold for the webhook endpoint should be high enough (or configurable) to accommodate legitimate multi-project monorepo setups where a single push can legitimately trigger several near-simultaneous webhook calls, or

A single repo-level webhook / batch endpoint should be available so that one incoming event can trigger reconciliation of all affected projects server-side, avoiding N separate HTTP calls entirely for N projects, or

At minimum, the rate limit configuration (threshold, window) should be exposed via an environment variable, so self-hosters with legitimate multi-project setups can tune it without disabling the protection entirely.

### Actual Behavior

A single push to the monorepo that touches multiple project subfolders (10+ configured webhooks near- simultaneously, regardless of which paths actually changed) results in some of the webhook-triggered syncs failing with {"error":"rate limit exceeded"}

### Screenshots

_No response_

### Arcane Version

2.8.0

### Installation Method

Docker Compose (Recommended)

### Environment Type

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

29.7.2

### Browser & Version

_No response_

```shell

```

```yaml

```

```shell

```

### Additional Context

No alternative exists in Arcane's UI to configure a single repo-level webhook that would internally reconcile all affected projects from one incoming event (this appears to be a duplicate/related concern to #1456, which requests a folder-scanning GitOps engine — commented there as well).

ohOgil:
## Checklist

- [x] This PR is **not** opened from my fork's `main` branch
- [x] All new user-facing strings are translated via Paraglide (`m.*()`)

## What This PR Implements

Webhook triggers (`/api/webhooks/trigger/:token`) were rate-limited by client IP.
When multiple webhooks are configured on the same Git host (e.g. Forgejo, GitHub),
they share a single token bucket, so a burst of legitimate triggers from different
webhooks on the same host can collide and return `429`, even though each webhook
targets a distinct project/container via its own token.

This PR changes the rate limiter to key on the webhook token instead, so each
webhook gets its own independent budget.

Fixes: #3652

## Changes Made

- Added `PerTokenRateLimitForPaths` in `rate_limit_middleware.go`, keying the
  rate limiter on the `:token` route param (SHA-256 hashed via the existing
  `agentTokenRateLimitKeyInternal` helper, consistent with the agent enrollment
  pattern) instead of the source IP.
- Rewired `/api/webhooks/trigger/:token` in `router_bootstrap.go` to use the new
  middleware. Per-token limits unchanged (60/min, burst 10).
- This mirrors how GitLab scopes webhook rate limiting by hook identity rather
  than IP, for the same reason: it isolates impact to the offending webhook
  instead of penalizing unrelated ones sharing a host.
- **Security fix from review:** the initial version let any well-formed-looking
  token get its own fresh bucket before validation, so an unauthenticated
  client cycling through arbitrary tokens could bypass rate limiting entirely
  and hammer the database-backed token lookup. Fixed with two additions in
  `PerTokenRateLimitForPaths`:
  - A wider IP-level ceiling (10x the per-token rate) checked first, always —
    bounds the cost of cycling through tokens regardless of shape.
  - `webhook.IsWellFormedToken` (shape check only — prefix, exact hex length,
    hex alphabet — no DB access) rejects structurally invalid tokens before
    any per-token bucket is allocated for them.
- Extracted a shared test helper (`newTokenRateLimitTestRouterInternal`) to
  reduce setup duplication across the `PerTokenRateLimitForPaths` tests.

## Testing Done

- Unit tests in `rate_limit_middleware_test.go`:
  - distinct tokens from the same IP are rate-limited independently
  - fallback to IP-based limiting when no token is present
  - IP ceiling bounds sustained cycling through unseen/invalid tokens
  - malformed tokens never get their own bucket and stay bounded by the IP ceiling
  - middleware only applies to configured paths
- 7 unit tests for `webhook.IsWellFormedToken` in `webhook_test.go` (valid
  shape, whitespace, wrong prefix, short/long hex, non-hex characters, empty).
- `go test ./internal/middleware/... ./internal/webhook/... ./internal/bootstrap/... -race -v`:
  all passing, no regressions.
- Manual verification via `curl` against the dev environment:
  - alternating 3 distinct tokens under load: all succeed independently, even
    while another token is rate-limited
  - single token burst (>10 rapid requests): correctly rate-limited (`429` +
    `Retry-After: 60`) after the burst, other tokens remain unaffected
  - malformed/invalid token: rejected `404`, no bucket allocated, no bypass
- Verified end-to-end against a real Forgejo instance triggering multiple
  webhooks in quick succession from the same source IP.

Claude was used to help diagnose the root cause, draft the
implementation and unit tests, and — after a P1 security issue was flagged
during automated review — help design and implement the fix (IP ceiling +
token shape validation). All code was reviewed, and all testing above (unit
tests, `go vet`, and manual `curl`/Forgejo verification) was performed
manually by me in a local dev environment before opening and updating this PR.

## Additional Context

This PR went through a security review round: the initial per-token keying
allowed unauthenticated clients to bypass rate limiting entirely by cycling
through arbitrary tokens (each got a fresh bucket before any validation).
Fixed by adding a token shape check and a wider IP-level ceiling checked
ahead of the per-token bucket. No configuration or environment variables
were added — thresholds are hardcoded as before, only the keying strategy
changed.

The PR changes public webhook trigger limiting from per-IP buckets to per-token buckets for stored tokens while retaining a shared per-IP limit for unknown tokens.
- Loads stored webhook token hashes into a synchronized in-memory index during startup.
- Keeps the index synchronized when webhooks are created or deleted.
- Adds middleware and service tests covering token isolation, unknown-token abuse limits, index loading, and concurrent CRUD behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

<sub>Reviews (11): Last reviewed commit: ["do at startup"](https://github.com/getarcaneapp/arcane/commit/2e4ba5bdd4c2eb4984e907ad4b13afea6aea5bc9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56750715)</sub>

**Context used:**

- Knowledge Base — [API services and request lifecycle](https://app.greptile.com/ofkm/-/custom-context/knowledge-base/getarcaneapp/arcane/-/docs/api-services.md)
- Knowledge Base — [API transport, realtime connections, and webhooks](https://app.greptile.com/ofkm/-/custom-context/knowledge-base/getarcaneapp/arcane/-/docs/api-transport-and-webhooks.md)


ohOgil:
@greptile review

backend/internal/middleware/rate_limit_middleware_test.go
```diff
@@ -162,6 +162,79 @@ func TestPerIPRateLimitForPaths_RouteParamsDoNotEscapeFilter(t *testing.T) {
 	require.Equal(t, http.StatusTooManyRequests, doReq("token-bbb"))
 }
 
+func TestPerTokenRateLimitForPaths_TracksDistinctTokensNotIP(t *testing.T) {
```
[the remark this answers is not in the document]
ohOgil: Fixed the duplication, but with a shared test helper instead of a single
table — these four tests check different scenarios (distinct tokens, missing
token, IP-ceiling bruteforce, malformed shape), so a table would blur what
each one proves. `newTokenRateLimitTestRouterInternal` now handles the
repeated router/request setup; each test only contains what's specific to it.

ohOgil:
@greptile review

ohOgil:
Thinking ahead: the current thresholds (60/min per token, burst 10; IP ceiling at 10x that) are hardcoded. Configurable **number of webhooks** isn't the concern here — each token gets its own independent bucket, so having many webhooks costs nothing on its own. The threshold that could eventually bite is trigger **rate**: a single busy webhook, or a large simultaneous burst across many webhooks sharing one IP.

For anyone running a larger number of webhooks than these defaults were tuned for, I'd like to follow up with a separate PR exposing these as env vars (something like `WEBHOOK_RATE_LIMIT_PER_MIN`, `WEBHOOK_RATE_LIMIT_BURST`, `WEBHOOK_RATE_LIMIT_IP_MULTIPLIER`), following the same pattern as `TRUSTED_PROXIES` in `config.go`. Keeping it out of this PR to stay focused on the security fix — will open a follow-up once this is merged.

ohOgil:
@greptile review

backend/internal/middleware/rate_limit_middleware.go
```diff
@@ -165,6 +165,137 @@ func agentTokenRateLimitKeyInternal(token string) string {
...
+// perTokenRateLimitIPMultiplierInternal widens the IP ceiling relative to
+// the per-token ceiling, since one IP can legitimately trigger many valid
+// tokens in a burst (e.g. a Git host firing several webhooks at once).
+const perTokenRateLimitIPMultiplierInternal = 10
+
+// knownValidTokenTTLInternal is how long a token that has already produced
+// a non-404 response is exempted from the shared IP ceiling. This keeps a
+// busy attacker cycling through invalid tokens from starving a legitimate,
+// already-confirmed webhook sharing the same source IP.
+const knownValidTokenTTLInternal = 10 * time.Minute
```
kmendell:
```suggestion
const (
	knownValidTokenTTLInternal = 10 * time.Minute
	perTokenRateLimitIPMultiplierInternal = 10
)
```

backend/internal/middleware/rate_limit_middleware.go
```diff
@@ -165,6 +165,137 @@ func agentTokenRateLimitKeyInternal(token string) string {
...
+
+// knownTokenCache tracks tokens that have already been confirmed to exist
+// (any response other than 404), so repeat requests for them can bypass the
+// shared IP ceiling and go straight to their own per-token bucket.
+type knownTokenCache struct {
```
kmendell: use samber/hot for caching instead of a raw mutex

backend/internal/middleware/rate_limit_middleware.go
```diff
@@ -165,6 +165,137 @@ func agentTokenRateLimitKeyInternal(token string) string {
...
+
+// PerTokenRateLimitForPaths rate-limits by client IP first (at
+// perTokenRateLimitIPMultiplierInternal times perMinute/burst), then by the
+// ":token" route param, for paths in the paths list (matched against
+// c.Path()). Once isValidResponse confirms a token via a successful
+// request, that token is cached and bypasses the IP ceiling on later
+// requests, so an attacker cycling through invalid tokens on a shared IP
+// cannot starve a legitimate, already-confirmed webhook. isValidShape, if
+// given, rejects malformed tokens before they get a bucket; both callbacks
+// must be cheap and side-effect free.
```
kmendell:
```suggestion
// PerTokenRateLimitForPaths rate-limits by client IP first (at
// perTokenRateLimitIPMultiplierInternal times perMinute/burst), then by the
// ":token" route param, for paths in the paths list (matched against
// c.Path()).
```

backend/internal/webhook/webhook.go
```diff
@@ -119,6 +119,19 @@ func parseWebhookPrefixInternal(raw string) (string, error) {
 	return webhookTokenPrefix + hexPart[:webhookTokenPrefixLen], nil
 }
 
+// IsWellFormedToken reports whether raw has the shape of a webhook token,
+// without a database lookup. A true result is not proof the token exists;
+// only TriggerByToken can determine that.
+func IsWellFormedToken(raw string) bool {
```
kmendell: im confused on the purpose of this or why it was needed? the webhook would also be allowed if permissions grants it and is looked up in the database, this just seems like un-needed logic.

kmendell (changes requested):

backend/internal/webhook/webhook.go
ohOgil: It's not doing authorization — `TriggerByToken` is still the sole source of truth for whether a token exists/is enabled. This is only a pre-filter for the rate limiter: without it, any random string hitting the trigger endpoint got its own fresh rate-limit bucket before ever reaching `TriggerByToken`, so an attacker could cycle arbitrary strings to bypass rate limiting entirely and hammer the DB lookup. `IsWellFormedToken` just avoids handing out a bucket to something that can't structurally be a real token (wrong prefix/length/hex), so that cost lands on the shared IP ceiling instead. Happy to drop it if you'd rather keep the rate limiter fully decoupled from the token format — the IP ceiling alone still bounds the abuse, just less precisely.

ohOgil:
@greptile review

backend/internal/middleware/rate_limit_middleware.go
```diff
@@ -165,6 +166,109 @@ func agentTokenRateLimitKeyInternal(token string) string {
...
+const (
+	// perTokenRateLimitIPMultiplierInternal widens the IP ceiling relative to
+	// the per-token ceiling, since one IP can legitimately trigger many valid
+	// tokens in a burst (e.g. a Git host firing several webhooks at once).
+	perTokenRateLimitIPMultiplierInternal = 10
+
+	// knownValidTokenTTLInternal is how long a confirmed-valid token is
+	// exempted from the shared IP ceiling, so it can't be starved by
+	// unrelated traffic on the same IP.
+	knownValidTokenTTLInternal = 10 * time.Minute
```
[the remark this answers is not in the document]
ohOgil: Confirmed — the cache had a re-entry gap: once a token's entry expired or got evicted, it had to clear the shared IP bucket again to get back in, which could stay blocked indefinitely under sustained unrelated traffic.

Planning to drop the cache and go back to the simpler two-layer design (shared IP ceiling checked first, then a per-token bucket for well-formed tokens) — no caching, so no re-entry gap. It's a known trade-off: a legitimate token can be delayed by noise on the same IP, but never permanently blocked. Will push shortly.

ohOgil:
@greptile review

ohOgil:
@greptile review

ohOgil:
Hi @kmendell, Greptile's check on this PR is showing red without any error detail, looks like it might be an issue on their end Let me know if you'd like me to retrigger it or if you're seeing something.

ohOgil:
Update : 
Greptile is right that this is a real trade-off, and I've gone back and forth on it (added a cache, then removed it after another finding showed the cache had its own starvation gap on expiry).

The two properties I haven't found a way to satisfy simultaneously with a simple design:
1. Bound token-cycling abuse → requires checking the shared IP ceiling first
2. Never let unrelated traffic on the same IP block an already-valid webhook
   → requires the token check to never depend on the shared IP budget

Current state picks (1) over (2): the IP ceiling is wide (10x the per-token rate) so it takes sustained abuse to trip, but a valid webhook can in theory be delayed if enough unrelated/malformed traffic shares its IP.

Given the back-and-forth, I'd rather get your take on which trade-off is acceptable here than keep guessing — happy to implement whichever direction you prefer, including reverting to the original per-IP-only limiting if the added complexity isn't worth it.

kmendell:
greptile doesnt produce comments if it is the same comments as previous runs regardless i pretty much fixed it 

the token body is already aes-gcm encrypted with the server key, so we can just decrypt it in the middleware if it decrypts, arcane generated it. those can skip the ip bucket and get limited per token, everything else stays ip limited. random tokens wont pass that and will still get limited

kmendell:
well i guess it slipped my mind that that is cpu intensive.

ohOgil:
Proposal: rather than reordering (which reintroduces the starvation issue), add a global rate limit specifically on decrypt attempts - a single shared token-bucket capping how many `isAuthentic` calls can run per second, independent of IP or token. This bounds total CPU spent on decryption server-wide without ever blocking a legitimate token based on what other IPs/tokens are doing. Something like:

```
    var decryptLimiter = rate.NewLimiter(rate.Limit(500), 1000) // tune to hardware

    if token != "" {
        if !decryptLimiter.Allow() {
            return 429 // decrypt capacity exhausted, try again shortly
        }
        if isAuthentic(token) { ... }
    }
```

kmendell:
@ben-kuhn You wouldn't happen to have ideas on this would you?

ohOgil:
Hi, on the Greptile P1 about `isAuthentic` running before the IP limiter: I looked into it and I don't think it's fixable without giving up what this PR is for.

To know a token is authentic (and let it bypass the shared IP ceiling), we have to decrypt it. So "no crypto once the IP bucket is empty" and "an authentic token is never starved by unrelated traffic from the same IP" can't both hold. I tried the reordering Greptile implies, and it fails the test this PR is built around:


```
--- FAIL: TestPerTokenRateLimitForPaths_AuthenticTokenNotStarvedByGarbageFromSameIP
        expected: 200, actual: 429
```

That's exactly the bug the PR fixes; webhooks dropped because something unrelated behind the same NAT/proxy/CI IP drained the bucket.

On cost, worst case is a forged token that clears the length gate and reaches the decrypt:

`BenchmarkIsAuthenticTokenForgedFullLength-4   1951612   1213 ns/op   3152 B/op   11 allocs/op`

~1.2 µs per rejected request, bounded to a single fixed-size decrypt by the length gate. The TLS handshake, HTTP parsing and the JSON 429 for that same request each cost more, so there's no amplification, the attacker always pays more than we do.

So I'd suggest keeping it as-is. Happy to add a short comment above the check documenting the ordering constraint, so it doesn't get re-flagged in future reviews.

@kmendell no rush at all, just keeping this on the radar since it's been quiet. If it's hard to get a second opinion on this one, I'm happy to go with whatever you decide, or to leave it open a bit longer if you'd rather wait.

kmendell (approved):

PR #3896, opened 2026-09-09
refactor(projects): carry environment ownership as groundwork for global views

mwd102:
Arcane seems to have split UX philosophy, with Dashboard presenting a global view of all environments but then updates/projects/container etc. pages are scoped to a single environment selected in the left-hand bar. It seems having a global environment view would be useful - allowing all resources e.g. containers, images, networks, to be visible in a single pane. This 'Global' option should live alongside current scoped behaviour as a selected option in the environment picker.
  
However, the main barrier to that is that objects do not currently carry their environmentId in views such as Projects (relying on the selected environment value instead).  This change attaches environmentId to project objects  when they are fetched and passes that environment explicitly through project actions, workspace requests, logs, and related container operations.
  
This change focused on projects only to demonstrate the pattern, if acceptable the natural evolution would be to continue across all relevant views such that a 'global' view can be made available in the environment picker that shows everything. 
  
This change is currently neutral, there is no added behaviour and no regression - all tests pass. User UI testing shows it behaves as expected. It is submitted for consideration and comments on overall approach and if this is or is not a direction Arcane should head in. 
  
  Validation

  - Frontend/backend builds, formatting, lint, and backend/CLI/shared-types tests passed.
  - E2E passing, CLI tests initially failed but passed when rerun with CI’s proxy configuration
  - Deployed and running in my own installation without issue.

  AI assistance

  Design approach was mine. OpenAI Codex assisted with implementation and validation.

### DIFF
--- a/.github/workflows/hyperion-agent-policy.yml
+++ b/.github/workflows/hyperion-agent-policy.yml
@@ -1,0 +1,24 @@
+name: Hyperion agent policy
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  hyperion-agent-policy:
+    name: Hyperion agent policy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Validate managed agent context
+        shell: bash
+        run: |
+          test "$(grep -cF '<!-- hyperion-context:start -->' AGENTS.md)" -eq 1
+          test "$(grep -cF '<!-- hyperion-context:end -->' AGENTS.md)" -eq 1
+          grep -qF 'globally configured Hyperion OKF memory' AGENTS.md
+          grep -qF '/home/homelab/Worktrees/<repo>/<task>' AGENTS.md
+          grep -qF 'Agents own the task Git lifecycle' AGENTS.md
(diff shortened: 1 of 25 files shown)


kmendell:
What you referring to is the small stesp ive been taking to get it to. place to show all resources, Ive been slowly trying to get it to a place where it can show everytihng however, the fatest way and most efficent way is going to be a push based agent, which will include breaking changes, i have a branch locally with a idea, but its taking me alot longer to work on it then i have free time.

kmendell:
Im open to see your idea, but just inclduing everything in the frontned is not up to par with what i would like arcane to be, As for all intensive purposes javascript is slow, while go is fast

mwd102:
Makes sense, I haven't looked at the agent side. The UI changes in this PR are more light touch than your vision and leverage the current backend API as is. Aim was to suggest an approach to unify the views around projects, containers, etc. but not require an updated agent resource model. I'm open to taking a look at that if you're willing to share the branch?

mwd102:
I took a quick look, I think I understand your concern - the api is pulling from every host on certain page loads which won't scale.

kmendell:
Like I said feel free to lay ground work for it, any help is welcome, just I want it to be seamless and just waiting on api calls

PR #3902, opened 2026-09-10
fix: handle Easy Join discovery on non-manager environments

### LINKED ISSUE #3897: 🐞 Bug: swarm/join-candidates returns 500 on edge environments when Swarm isn't used
### Bug Description

I use Arcane 2.10.2 to manage 5 hosts, 1 local plus 4 edge agents. None of them run Swarm, they're all plain Docker with compose files.

The UI polls `/api/environments/{id}/swarm/join-candidates` every 5 seconds. On the four edge environments it always returns 500. The toast fires constantly, and on Settings pages the whole page gets replaced with "Cannot connect to local Docker service", which is misleading because Docker is perfectly fine.

One thing I noticed: the local environment (id 0) returns 409, not 500. I suspect that's why this hasn't been reported much, single host installs won't hit it.

It fails in 10 to 50ms, so it isn't timing out waiting on the agents. Using an admin API key makes no difference, still 500.

**The permission side**

For users who don't have `swarm:join` on every environment, the same endpoint returns 403 instead.

The docs say Easy Join needs `swarm:join` on every target. In our setup the team leads have Admin on dev/test/prod but read only (Monitor) on a GPU box, so that single read only environment makes the call fail everywhere for them. Since it's a background poll rather than something they clicked, the 403 just shows up as a mysterious "permission denied: swarm:join" toast with no context.

So depending on the user it's either 403 or 500, but the result is the same, an error every 5 seconds.

**Workaround**

For now I intercept that one path in nginx and return `{"success":true,"data":[]}`. Since we don't use Swarm an empty list is the correct answer anyway, but obviously I'd rather drop the hack.

Possibly related to #3828, which is also Easy Join failing with a 500 and no output, though that one is about actually joining rather than listing candidates.

### Steps To Reproduce

1. Run a manager with at least one edge agent environment added (I have four)
2. Leave Swarm uninitialised on every host, plain Docker only, so `swarm/status` reports `{"enabled": false}`
3. Open the UI with an edge environment selected and leave it on any page
4. Watch the network tab, `/api/environments/{id}/swarm/join-candidates` is polled every 5 seconds and returns 500 each time
5. Go to Settings, the page renders "Cannot connect to local Docker service" instead of loading

For the 403 variant, do the same with a user who has read only access on at least one environment.

### Expected Behavior

When there is no Swarm manager to join, I'd expect an empty candidate list rather than an error. Alternatively the UI could skip polling Easy Join entirely when `swarm/status` reports `enabled: false`.

It would also help if the UI didn't render every 500 as "cannot connect to local Docker service". That message sent me digging through the Docker socket and the agents for a while before I realised it was one unrelated endpoint failing.

### Actual Behavior

The endpoint returns 500 on every edge environment, and because the UI polls it, that happens roughly every 5 seconds for as long as the tab is open.

On Settings pages the entire page is replaced with "Cannot connect to local Docker service", even though Docker is running normally and every other API call succeeds at the same moment. `/api/users`, `/api/roles`, `/api/environments` and `swarm/status` all return 200 while this one fails, so the page has the data it needs but renders the error instead.

For users who don't have `swarm:join` on every environment it returns 403 rather than 500, which surfaces as a "permission denied: swarm:join" toast. Since nothing was clicked to trigger it, there's no indication of which action was denied or why.

### Screenshots

_No response_

### Arcane Version

2.10.2

### Installation Method

Docker Compose (Recommended)

### Environment Type

Remote Agent Mode

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

28.5.2

### Browser & Version

Edge 151.0.4129.59

```shell

```

```yaml

```

```shell

```

### Additional Context

_No response_

kmendell:
## Checklist

## What This PR Implements

Fixes: https://github.com/getarcaneapp/arcane/issues/3897

## Additional Context

The PR updates Easy Join discovery so candidate lookup works when the selected environment is not itself an active manager, moves frontend discovery to an environment- and permission-scoped Svelte Query, propagates request cancellation, improves remote error mapping, and tightens connection-error presentation.
- Validates the selected manager identity before discovering eligible environments.
- Filters candidates by existing node bindings and user permissions.
- Uses a scoped query key and abort signal to protect environment switches from stale responses.
- Handles proxied HTTP status errors and absent remote identity data explicitly.
</details>

The PR appears safe to merge, with no outstanding blocking or actionable new issue identified.

The current query implementation scopes results to the selected environment and permission state and cancels obsolete requests, while the backend validates manager identity before discovery. Both previous findings were manually resolved without explanatory replies and therefore are not outstanding.
</details>

<sub>Reviews (2): Last reviewed commit: ["fix: handle Easy Join discovery on non-m..."](https://github.com/getarcaneapp/arcane/commit/f644554602b8f563e0c7b33b243afd91fdd06115) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62766780)</sub>

**Context used:**

- Knowledge Base — [Swarm orchestration and cluster management](https://app.greptile.com/ofkm/-/custom-context/knowledge-base/getarcaneapp/arcane/-/docs/swarm-orchestration.md)
- Knowledge Base — [Web application experience](https://app.greptile.com/ofkm/-/custom-context/knowledge-base/getarcaneapp/arcane/-/docs/web-application.md)

### DIFF
--- a/backend/internal/oidc/service.go
+++ b/backend/internal/oidc/service.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/samber/hot"
-	"golang.org/x/net/http2"
 	"golang.org/x/oauth2"
 
 	"github.com/getarcaneapp/arcane/backend/v2/internal/config"
@@ -140,11 +139,12 @@ func (s *OidcService) getInsecureHttpClientInternal() *http.Client {
 	} else {
 		insecureTransport.TLSClientConfig.InsecureSkipVerify = true
 	}
-	// Force HTTP/2 even with custom TLS config to avoid "malformed HTTP response" errors
-	// when the server speaks HTTP/2 but the client disabled it due to custom TLS config.
-	if err := http2.ConfigureTransport(insecureTransport); err != nil {
-		slog.Warn("getInsecureHttpClientInternal: failed to configure http2 transport", "error", err)
+	// Enable HTTP/2 even with a custom TLS configuration.
+	if insecureTransport.Protocols == nil {
+		insecureTransport.Protocols = new(http.Protocols)
+		insecureTransport.Protocols.SetHTTP1(true)
 	}
+	insecureTransport.Protocols.SetHTTP2(true)
 	insecureClient.Transport = insecureTransport
 	s.insecureHttpClient = &insecureClient
 	return s.insecureHttpClient
--- a/backend/internal/swarm/easy_join.go
+++ b/backend/internal/swarm/easy_join.go
@@ -1,0 +1,66 @@
+package swarm
+
+import (
+	"context"
+
+	"emperror.dev/errors"
+	"github.com/getarcaneapp/arcane/backend/v2/internal/environment"
+	"github.com/getarcaneapp/arcane/backend/v2/pkg/pagination"
+	swarmtypes "github.com/getarcaneapp/arcane/types/v2/swarm"
+)
+
+// GetJoinCandidates lists environments available to join the selected manager.
+func (s *SwarmService) GetJoinCandidates(ctx context.Context, environmentID string) ([]swarmtypes.SwarmJoinCandidate, error) {
+	var identity *SwarmNodeIdentity
+	var err error
+	if environmentID == environment.LocalEnvironmentID {
+		identity, err = s.GetLocalNodeIdentity(ctx)
+	} else {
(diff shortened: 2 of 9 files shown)


kmendell:
* **#3902** <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3902?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getarcaneapp/arcane/3902?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1861, opened 2026-02-23
feat: editor enhancements, switch back to code mirror editor

kmendell:
## What This PR Implements

**Related issue**

## Related Issue

Fixes # https://github.com/getarcaneapp/arcane/issues/1365
Fixes # https://github.com/getarcaneapp/arcane/issues/1368

## Changes Made

-
-
-

## Testing Done

- [ ] Development environment started: `./scripts/development/dev.sh start`
- [ ] Frontend verified at http://localhost:3000
- [ ] Backend verified at http://localhost:3552
- [ ] Manual testing completed (describe):
- [ ] No linting errors (e.g., `just lint all`)
- [ ] Backend tests pass: `just test backend`

## Checklist

- [ ] This PR is **not** opened from my fork’s `main` branch

## AI Tool Used (if applicable)

AI Tool:
Assistance Level: 
What AI helped with:
I reviewed and edited all AI-generated output:
I ran all required tests and manually verified changes:

## Additional Context

**Disclaimer** Greptiles Reviews use AI, make sure to check over its work.

To better help train Greptile on our codebase, if the comment is useful and valid **Like** the comment, if its not helpful or invalid **Dislike**

<details open><summary><h3>Greptile Summary</h3></summary>

This PR migrates the code editor from Monaco to CodeMirror with significant enhancements to validation and user experience.

**Major Changes:**
- Replaced Monaco editor with CodeMirror editor implementation
- Added comprehensive Docker Compose YAML schema validation using AJV
- Implemented environment variable analysis with duplicate detection and secret warnings
- Added autocomplete for YAML compose files and env files with context-aware suggestions
- Introduced diff/merge view for comparing changes
- Added outline panel for navigating YAML structure and env variables
- Implemented command palette for editor actions
- Enhanced validation prevents saving/creating projects when syntax errors exist
- Updated all editor usages across projects, templates, and new project pages
- Updated Playwright tests to work with CodeMirror selectors

**Quality Improvements:**
- Better mobile editor support (mentioned in recent commit history)
- Proper validation state management with `hasErrors` and `validationReady` flags
- Context-aware variable resolution across compose and env files
- Improved user feedback with diagnostic summaries showing error/warning counts
</details>

<details open><summary><h3>Confidence Score: 5/5</h3></summary>

- This PR is safe to merge with minimal risk
[...]

PR #3985, opened 2026-09-16
fix: nest services/networks under their section in compose outline

### LINKED ISSUE #3981: 🐞 Bug: Compose editor Outline panel never shows individual services/networks nested under their section
### Bug Description

The compose editor's Outline panel is supposed to show a second level of detail — each individual service, network, volume, config, and secret name nested under its section header — but in practice it only ever renders the flat section headers (`services`, `networks`, etc.) themselves. This makes the Outline panel far less useful on any stack with more than one or two services, since there's no way to jump straight to a specific service.

### Steps To Reproduce

1. Open any Project's Configuration tab (compose editor) — works with any compose file that has at least one service.
2. Click the Outline toggle button in the editor toolbar.
3. Observe the Outline panel that opens.

### Expected Behavior

Per the existing `buildOutline()` logic in `compose-analysis.ts`, each section (`services`, `networks`, `volumes`, `configs`, `secrets`) should get a top-level entry, and every key under that section (e.g. each service name) should appear as an indented, clickable child entry that jumps to that service's location in the file.

### Actual Behavior

Only the flat section headers ever appear. No service/network/volume/config/secret names show up underneath them, regardless of how many are defined in the file.

**Root cause:** `buildOutline()` calls `Object.keys()` on the value returned by `doc.getIn([section])` (from the `yaml` npm package), expecting a plain object. That value is actually a `YAMLMap` class instance, so `Object.keys()` returns the instance's own properties (`items`, `range`) instead of the real YAML keys. Every subsequent lookup for those bogus "keys" resolves to nothing, so the child entries get silently dropped. Reproduced locally with the `yaml` package directly to confirm.

### Arcane Version

v2.12.0 (also reproduces on current `main`)

### Installation Method

Docker Compose

### Environment Type

### Database Type

SQLite (Default)

### Operating System

Linux

### Docker Version

29.0.4

### Browser & Version

Chrome 153.0.8010.36 (Official Build) (x86_64)

### Additional Context

A fix is ready and tested against a live multi-service stack in a dev build; will open a PR referencing this issue.

angst911:
## Checklist

- [x] This PR is **not** opened from my fork's `main` branch
- [x] All new user-facing strings are translated via Paraglide (`m.*()`) — n/a, no new strings

## What This PR Implements

Fixes #3981

The compose editor's Outline panel shows section headers (`services`, `networks`, etc.) but never the entries nested under them, even though the code was clearly written to do that. Present since the commit that introduced `buildOutline()` (#1861) — not a later regression.

`buildOutline()` called `Object.keys()` on a `YAMLMap` instance from the `yaml` package instead of the map's actual entries — reading off the class's own internal properties (`items`, `range`) instead of real YAML keys. Every lookup failed silently, so nested items never got added.

## Changes Made

- `buildOutline()` now iterates the section's `YAMLMap.items` directly (using the `isMap`/`isPair`/`scalarToKey` helpers already used elsewhere in this file) instead of calling `Object.keys()` on the resolved value.
- No changes needed in `editor.svelte` or `types.ts` — the outline UI already rendered `level: 1` items correctly, it just never received any.

## Testing Done

Ran the dev stack against a real multi-service compose file (8 services, several networks), not a throwaway test project. I (a human, not the AI assistant) logged into the dev instance and checked the Outline panel in the browser myself, before and after: before, only section headers appeared; after, every service/network shows up nested under its section, and clicking one jumps to the right line. `vp check --no-fmt` passes clean on the changed file.

Scope: confirmed Arcane started, read the compose file, and the Outline panel worked correctly. No broader regression pass over the rest of the editor (schema validation, diagnostics, diffing) or other Arcane features — this change is scoped to `buildOutline()` only.

Used Claude Code to investigate, fix, and test this. It found the root cause by reproducing the `Object.keys()`-on-`YAMLMap` behavior locally with the `yaml` package, wrote the fix, and set up the dev environment — but the functional verification in the browser was done by me, not the agent.

## Additional Context

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details open><summary>Greptile Summary</summary>

This PR corrects Compose outline generation by traversing each parsed YAML section map directly instead of enumerating the `YAMLMap` object’s implementation properties.
- Uses parsed node ranges consistently for section and nested outline items.
- Adds service, network, volume, config, and secret entries beneath their corresponding section headers.
- Reuses the analyzer’s existing YAML node helpers.
</details>

<details open><summary>Confidence Score: 5/5</summary>

The PR appears safe to merge, with no concrete correctness, security, or repository-rule issues identified.

The changed traversal correctly operates on parsed `YAMLMap` pairs, filters unsupported nodes using existing helpers, and preserves valid source ranges for generated outline entries.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: nest services/networks under their ..."](https://github.com/getarcaneapp/arcane/commit/a904f23eadfd6120e31820e4a277f1aef67c2522) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=65107957)</sub>


kmendell (approved):
Thanks!

